PR #5, opened 2020-12-09
Adding Power support(ppc64le) with ci and testing to the project for architecture independent

asellappen:
I am working for IBM to port cpu arch ppc64le for open sources.

This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro 
on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.,For more info tag @gerrith3.

please note mypy is failing in both cpu arch amd64 and ppc64le ,.
Please help to verify and merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ language: python
 cache:
   directories:
     - $HOME/.cache/pip
-
+arch:
+ - amd64
+ - ppc64le
 python:
   - "3.6"
   - "3.7"
@@ -52,6 +54,34 @@ jobs:
     - stage: test
       name: mypy
       python: "3.6"
+      install: python3 -m pip install mypy .
+      script: mypy src/
+  
+    - stage: deploy
+      arch: ppc64le
+      services:
+        - docker
+      python: "3.6"
+      install: python3 -m pip install Cython twine
+      if: tag IS present
+      script:
+        - |
+          python3 setup.py sdist
+          ./buildwheels.sh
+          ls -l dist/
+          python3 -m twine upload dist/*
+
+    - stage: test
+      name: flake8
+      arch: ppc64le
+      python: "3.6"
+      install: python3 -m pip install flake8
+      script: flake8 src/ tests/
+
+    - stage: test
+      name: mypy
+      python: "3.6"
+      arch: ppc64le
       install: python3 -m pip install mypy .
       script: mypy src/
 


marcelm:
Hi, I understand why this would be interesting for IBM. However, I am going to close this PR for the following reasons. 

* This PR adds a lot of duplication. For example, there is no need to run `flake8` again on `ppc64le` because it should give the same results on any architecture.
* This adds time to CI tests. Some extra time is ok, but I would rather use it for tests on macOS, Windows or ARM before adding PowerPC.
* I don’t have the knowledge to fix issues that arise if a test fails on `ppc64le`.
* I have never heard of anyone using `dnaio` on `ppc64le` (nor Cutadapt, which has `dnaio` as a dependency). This either means that it is not used on that architecture very much or that it works without problems, and that therefore the existing tests are sufficient.
* I am going to switch to GitHub actions soon as Travis has it made harder for Open Source projects to freely use their service.

To be honest, if IBM has an interest in this package working on their architecture, they should do the testing themselves instead of externalizing the costs for doing so to volunteers.

I am happy to accept PRs that fix platfom-specific problems in `dnaio`.

asellappen:
Thanks @marcelm for looking at this PR. Please find my inline reply for your reasons to close PR ,This is just to clarify you.  Please tag @gerrith3 for more info.

> Hi, I understand why this would be interesting for IBM. However, I am going to close this PR for the following reasons.
> 
> * This PR adds a lot of duplication. For example, there is no need to run `flake8` again on `ppc64le` because it should give the same results on any architecture.

flake8 toolkit for checking our code base against coding style (PEP8), programming errors and other qualties,If there is any issue based on cpu arch then this test may help to identify and fix, In case if you think the code is generic and will not be there any issue based on arch then i am fine to remove this part from package.

> * This adds time to CI tests. Some extra time is ok, but I would rather use it for tests on macOS, Windows or ARM before adding PowerPC.

Agreed,  However we are taking care of PowerPC and we want to ensure ,This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and ISVs, and while we don't use this package directly, we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or third level dependency and couldn't tell you explicitly which features we use or our usage model.
> * I don’t have the knowledge to fix issues that arise if a test fails on `ppc64le`.
We are there always to help and fix the issues for ppc64le.
> * I have never heard of anyone using `dnaio` on `ppc64le` (nor Cutadapt, which has `dnaio` as a dependency). This either means that it is not used on that architecture very much or that it works without problems, and that therefore the existing tests are sufficient.

As i said earlier there , we don't use this package directly, we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or third level dependency and couldn't tell you explicitly which features we use or our usage model.


> * I am going to switch to GitHub actions soon as Travis has it made harder for Open Source projects to freely use their service.

you can get free credit for CI time on travis ,To request for more build minutes, create a request at https://support.travis-ci.com/hc/en-us

PR #13, opened 2021-09-27
Add more explanation about the decode('latin-1') speed advantage

rhpvorderman:
I checked if `.decode('ascii')` could be used as a shorthand to check whether these strings are valid ASCII. It is possible but it comes at a speed disadvantage

```
  Time (mean ± σ):      1.958 s ±  0.050 s    [User: 1.843 s, System: 0.114 s]
  Range (min … max):    1.918 s …  2.094 s    10 runs

  Time (mean ± σ):      2.311 s ±  0.076 s    [User: 2.196 s, System: 0.115 s]
  Range (min … max):    2.256 s …  2.507 s    10 runs
 
```
I dived deeper in the Python code and put my findings in the comments as to why this is.


rhpvorderman:
Incidentally this means corners can be cut by directly using the internal string pointer without checking. But this is very dangerous for user-generated Sequence objects.

PR #14, opened 2021-12-09
Improve parser performance

rhpvorderman:
Hi. I have been looking at this code for more than 3 years, and finally I find a way to improve the performance. This consists of two commits, of which you may choose to merge only one.

### Base performance
```
  Time (mean ± σ):      2.080 s ±  0.018 s    [User: 1.956 s, System: 0.123 s]
  Range (min … max):    2.055 s …  2.105 s    10 runs
```

### Improve code layout

Commit one groups similar functionality in the parse code. This makes the code easier to read. As well as limits the amount of '+1', '-1', etc. actions. We don't create new name, sequence and qualities variables until it is certain that the FASTQ record has the complete 4 newlines.
This code is slightly faster (5-10% in my benchmarks). Maybe this has also to do with the branch predictor in the CPU having an easier time because similar functionality is grouped. (EDIT: Or this may be the layout issue mentioned below in which case it is a very small increment probably)
```
  Time (mean ± σ):      1.878 s ±  0.014 s    [User: 1.769 s, System: 0.109 s]
  Range (min … max):    1.855 s …  1.907 s    10 runs
```

### Infer quality length from sequence length

I was reading [joelonsoftware.com's article on microsoft office formats](https://www.joelonsoftware.com/2008/02/19/why-are-the-microsoft-office-file-formats-so-complicated-and-some-workarounds/), and he noted that binary formats are much easier to read, because you can simply copy whole parts in memory directly. That is also what happens in this parser, which is good. I was lamenting that it is unfortunate that we do not know the length beforehand because FASTQ does not encode that.

Then I realized the spec says that the qualities string MUST be the length of the sequence string. So it **does** encode one length.

For each character in the input, two checks were performed. Is it a newline, and have we reached bufend yet?
With this change the length of the quality string is inferred from the sequence string. Therefore we can skip
sequence length x 2 checks. We just perform checks afterward to see if we end on a newline.

This has a slight risk, as there are some possible outcomes:
1. The newline is indeed the last newline of the record.
2. The newline is a newline in a following record, because the quality string was truncated and we were unlucky.

In case of 1. Success, but in case of 2. there are two possible outcomes:
a. It is the ending newline of the name, sequence or second header line.
   In which case the parser will throw an error for the next record.
   Which is good, problem averted.
b. It is the ending newline of the next record's quality string. In that case
   we have included an entire record in this record's quality string and the
   parser keeps happily parsing.

In case of b, this leads to a missing record, and a record with
invalid quality values. This might be catched by upstream tools
(problem averted!) or even by cutadapt itself (in paired mode), or it will lead to wrong quality values for this record.
Probably one misrepresented record will not lead to a different analysis outcome.
 If there are multiple records with an erroneous quality length, they are bound to be not lucky all the time and trigger parser erros.

So with this commit we sacrifice fidelty in an extremely rare edge case error for speed.  Dnaio does this all the time though: quality and sequence are not checked for if the characters are in the valid ASCII range. So this is in the spirit of the project. EDIT: to better phrase this. Garbage in == garbage out. With this code change the ability is lost to catch one very rare case of garbage. With correct input the output will still always be correct.

And now the speed improvement:

```
  Time (mean ± σ):      1.310 s ±  0.017 s    [User: 1.182 s, System: 0.128 s]
  Range (min … max):    1.278 s …  1.338 s    10 runs
``` 

### Disclaimer

Compiling C is prone to layout problems which can make the performance differ as much as 40%. See this very interesting video below. I did not benchmark this change with the rigor as shown in this video. However because the code changes do demonstrably less work each, I think the performance gains are real.
https://www.youtube.com/watch?v=7g1Acy5eGbE

EDIT: After deleting two unused variables for instance after recompilation the speed improvement is:
```
  Time (mean ± σ):      1.462 s ±  0.019 s    [User: 1.331 s, System: 0.130 s]
  Range (min … max):    1.438 s …  1.502 s    10 runs
```
Still good, but a 150 ms difference!  For compiler and layout things that are not easy to fix.

### DIFF
--- a/tests/test_internal.py
+++ b/tests/test_internal.py
@@ -239,7 +239,7 @@ class TestFastqReader:
             list(fq)
 
     def test_second_header_not_equal(self):
-        fastq = BytesIO(b'@r1\nACG\n+xy\n')
+        fastq = BytesIO(b'@r1\nACG\n+xy\nXXX\n')
         with raises(FastqFormatError) as info:
             with FastqReader(fastq) as fq:
                 list(fq)  # pragma: no cover
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -212,7 +212,9 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
         int endskip
         str name
         char* name_encoded
-        Py_ssize_t bufstart, bufend, pos, record_start, sequence_start
+        Py_ssize_t last_read_position = 0
+        Py_ssize_t record_start = 0
+        Py_ssize_t bufstart, bufend, sequence_start
         Py_ssize_t second_header_start, sequence_length, qualities_start
         Py_ssize_t second_header_length, name_length
         Py_ssize_t name_start, name_end, second_header_end, sequence_end
@@ -259,11 +261,20 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
                 bufstart += 1
                 bufend += 1
                 extra_newline = True
-            else:
-                break
+            elif last_read_position > record_start:  # Incomplete fastq records are present.
+                    if extra_newline:
+                        last_read_position -= 1
+                    lines = buf[record_start:last_read_position].count(b'\n')
+                    raise FastqFormatError(
+                        'Premature end of file encountered. The incomplete final record was: '
+                        '{!r}'.format(
+                            shorten(buf[record_start:last_read_position].decode('latin-1'),
+                                    500)),
+                        line=n_records * 4 + lines)
+            else:  # EOF Reached. Stop iterating.
+                return
 
         # Parse all complete FASTQ records in this chunk
-        pos = 0
         record_start = 0
         while True:
             ### Check for a complete record (i.e 4 newlines are present)
@@ -365,7 +376,7 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
             n_records += 1
             record_start = next_record_start
         # bufend reached
-        pos = bufend
+        last_read_position = bufend
         if record_start == 0 and bufend == len(buf):
             # buffer too small, double it
             buffer_size *= 2
@@ -379,14 +390,6 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
         else:
             bufstart = bufend - record_start
             buf[0:bufstart] = buf[record_start:bufend]
-    if pos > record_start:
-        if extra_newline:
-            pos -= 1
-        lines = buf[record_start:pos].count(b'\n')
-        raise FastqFormatError(
-            'Premature end of file encountered. The incomplete final record was: '
-            '{!r}'.format(shorten(buf[record_start:pos].decode('latin-1'), 500)),
-            line=n_records * 4 + lines)
 
 
 def record_names_match(header1: str, header2: str):
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -269,46 +269,35 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
             ### Check for a complete record (i.e 4 newlines are present)
             # Use libc memchr, as some libc functions will be implemented in assembly
             # void *memchr(const void *str, int c, size_t n)
-            name_end_ptr = <char *>memchr(c_buf + pos, 10, <size_t>(bufend - pos))
+            name_end_ptr = <char *>memchr(c_buf + record_start, 10, <size_t>(bufend - record_start))
             if name_end_ptr == NULL or name_end_ptr == bufend_ptr:
-                pos = bufend
                 break
             name_end = name_end_ptr - c_buf
-            pos = name_end
-            pos += 1
-            sequence_start = pos
-            sequence_end_ptr = <char *>memchr(c_buf + pos, 10, <size_t>(bufend - pos))
+            sequence_start = name_end + 1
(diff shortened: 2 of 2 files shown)


rhpvorderman:
@marcelm You will **love** this. Instead of using C code to find the newlines I used `memchr`. A lot of the `mem*` code in glibc is implemented in assembler.

And there we go:
```
  Time (mean ± σ):      1.069 s ±  0.013 s    [User: 946.3 ms, System: 122.3 ms]
  Range (min … max):    1.047 s …  1.093 s    10 runs
```

marcelm:
Oh, really cool! I would love to look at this right now, but that will probably have to wait until next week, just so you know.

rhpvorderman:
No problem! I also checked out the memchr improvement while not using the quality length hack:
```
# No quality skip but with memchr
  Time (mean ± σ):      1.139 s ±  0.014 s    [User: 1.028 s, System: 0.111 s]
  Range (min … max):    1.125 s …  1.173 s    10 runs
```
So the performance savings are not as big when using memchr. It might be considered not doing it. I will make an alternative PR for this.

rhpvorderman:
I was thinking of a couple of ways to improve the parser still, but I thought it would be instructive to have a baseline benchmark for performance.

So I wrote this script:
```python3
#!/usr/bin/env python3

import sys

with open(sys.argv[1], "rb")  as file_h:
    for line in file_h:
        pass
```
This does almost exactly the same thing as dnaio. It reads in chunks of files and finds the line ends using memchr. It then copies the found lines directly into python objects. (It skips the encoding part, but we have already established that encoding with latin-1 is the same as copying it directly into a python object).
This creates exactly the same amount of python objects per FASTQ record (4 bytes objects vs 3 unicode objects and 1 custom class). The mechanisms are also mostly the same as CPython uses memchr internally.

The results:
```
  Time (mean ± σ):      1.104 s ±  0.025 s    [User: 997.1 ms, System: 106.9 ms]
  Range (min … max):    1.069 s …  1.158 s    10 runs

Benchmark #1: python binary_read.py ~/test/big2.fastq
  Time (mean ± σ):      1.004 s ±  0.038 s    [User: 872.2 ms, System: 132.1 ms]
  Range (min … max):    0.974 s …  1.085 s    10 runs
```

So the iteration of dnaio as presented in this PR is only 10% slower than the baseline performance. It will be impossible to get it below 1 second using python's file reading mechanisms. Optimizations to the parser code are moot after this point.
Significant gains can only be made by improving the file reading. Using mmaps is a possibility, but since we use gzipped fastq files almost exclusively in production, that will not work. And unpacking first and then parsing is slower anyway. (And a huge disk IO burner).

marcelm:
I’m reviewing this commit by commit, please let me know beforehand if you want to do any force-pushing.

First commit looks fine, nice improvement.

tests/test_internal.py
```diff
@@ -239,7 +239,7 @@ def test_two_header_detection(self):
             list(fq)
 
     def test_second_header_not_equal(self):
-        fastq = BytesIO(b'@r1\nACG\n+xy\n')
+        fastq = BytesIO(b'@r1\nACG\n+xy\nXXX\n')
```
marcelm, line 242: Can you please duplicate the test and test it with and without the trailing `XXX`?

rhpvorderman: I first check if the quality length in order, because I do a shortcut there. But that is not really necessary. I can revert the change to this test if I do the checks in the correct order again. Will do so after the entire review.

rhpvorderman: I Looked a bit at it. Removing the XXX quality line will give a Premature End Of File error. Which isn't wrong in this case. 
Due to the code reorganization a check for a comple fastq record is done first before any other checking or parsing is done. This is optimal, because reaching the end of the buffer happens quite often, and we want to refill the buffer ASAP instead of checking a FASTQ that is not complete anyway.

rhpvorderman:
If it helps the review process. #15 is basically the improvement of the parser using `memchr`. The only difference with this branch is this commit:
https://github.com/rhpvorderman/dnaio/commit/3c0edf8dffee65d49b4444d083c26e998c93d5a6

This commit highlights how I implemented the quality length shortcut.

tests/test_internal.py
marcelm: I don’t quite understand. I just want the existing test to work as before. If you want, you can of course add another test, but don’t change this one.

marcelm:
> EDIT: Although, of course I am in favour for not leaving the performance on the table. But I do not want to force the quality length shortcut and sugarcoat it with performance that is not derived from it.

I’ve thought about this a bit, and my tendency at the moment is to not include this shortcut. I think it’s smart and of course it makes things a bit faster, but it’s still a hack. We’re still working with Python, so there’s already a lot of performance left on the table, as you say, so I would tend to try to optimize other parts of the pipeline (of which dnaio is likely only a small part), instead.

rhpvorderman:
Yes, it seems to save only 80ms on 5 million illumina records, so that is not very stellar.

>  (of which dnaio is likely only a small part),

I am currently working on a fastq-filter program, written in Python + Cython, and dnaio is the vast majority of the compute time (80-90%). Hence my improvements here.

Could you have a look at #15 instead? That improves the speed by almost 100% by using memchr. It also reorganizes the fastq_iter code a bit to be more readable. It is code without shorcuts.
Thank you for your feedback so far.

marcelm:
> Could you have a look at #15 instead?

Yes of course! The rest is fine and unless I find something strange, I want to merge it.

PR #15, opened 2021-12-09
Improve parser speed by looking for newlines with memchr

rhpvorderman:
This is the alternative to #14.

EDIT: This commits can be squashed if need be.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -225,10 +225,10 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
         Py_ssize_t sequence_start, sequence_end, sequence_length
         Py_ssize_t second_header_start, second_header_end, second_header_length
         Py_ssize_t qualities_start, qualities_end, qualities_length
-        cdef char *name_end_ptr
-        cdef char *sequence_end_ptr
-        cdef char *second_header_end_ptr
-        cdef char *qualities_end_ptr
+        char *name_end_ptr
+        char *sequence_end_ptr
+        char *second_header_end_ptr
+        char *qualities_end_ptr
         bint custom_class = sequence_class is not Sequence
         Py_ssize_t n_records = 0
         bint extra_newline = False
@@ -289,29 +289,25 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
             # using 64-bit integers. See:
             # https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=string/memchr.c;hb=HEAD
             # void *memchr(const void *str, int c, size_t n)
-            name_end_ptr = <char *>memchr(c_buf + record_start, 10, <size_t>(bufend - record_start))
+            name_end_ptr = <char *>memchr(c_buf + record_start, b'\n', <size_t>(bufend - record_start))
             if name_end_ptr == NULL:
                 break
-            # The maximum value of name_end_ptr = c_buf + bufend - 1.
-            # c_buf + bufend points at the terminating null byte.
-            # c_buf + bufend - 1 is the last character in the buffer. Therefore
-            # we can always safely increase the pointer with +1 to look for the next
-            # newline. Because in case of the end of buffer, name_end will be
-            # bufend - 1. So sequence_start will be bufend. The search
-            # length will be bufend - bufend == 0 and memchr will return NULL.
+            # bufend - sequence_start is always nonnegative:
+            # - name_end is at most bufend - 1
+            # - thus sequence_start is at most bufend
             name_end = name_end_ptr - c_buf
             sequence_start = name_end + 1
-            sequence_end_ptr = <char *>memchr(c_buf + sequence_start, 10, <size_t>(bufend - sequence_start))
+            sequence_end_ptr = <char *>memchr(c_buf + sequence_start, b'\n', <size_t>(bufend - sequence_start))
             if sequence_end_ptr == NULL:
                 break
             sequence_end = sequence_end_ptr - c_buf
             second_header_start = sequence_end + 1
-            second_header_end_ptr = <char *>memchr(c_buf + second_header_start, 10, <size_t>(bufend - second_header_start))
+            second_header_end_ptr = <char *>memchr(c_buf + second_header_start, b'\n', <size_t>(bufend - second_header_start))
             if second_header_end_ptr == NULL:
                 break
             second_header_end = second_header_end_ptr - c_buf
             qualities_start = second_header_end + 1
-            qualities_end_ptr = <char *>memchr(c_buf + qualities_start, 10, <size_t>(bufend - qualities_start))
+            qualities_end_ptr = <char *>memchr(c_buf + qualities_start, b'\n', <size_t>(bufend - qualities_start))
             if qualities_end_ptr == NULL:
                 break
             qualities_end = qualities_end_ptr - c_buf
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -212,7 +212,9 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
         int endskip
         str name
         char* name_encoded
-        Py_ssize_t bufstart, bufend, pos, record_start, sequence_start
+        Py_ssize_t last_read_position = 0
+        Py_ssize_t record_start = 0
+        Py_ssize_t bufstart, bufend, sequence_start
         Py_ssize_t second_header_start, sequence_length, qualities_start
         Py_ssize_t second_header_length, name_length
         Py_ssize_t name_start, name_end, second_header_end, sequence_end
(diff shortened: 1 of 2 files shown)


src/dnaio/_core.pyx
```diff
@@ -258,122 +266,127 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
-            pos += 1
-
-            # Parse the sequence (line 1)
-            sequence_start = pos
-            while pos < bufend and c_buf[pos] != b'\n':
-                pos += 1
-            if pos == bufend:
+            name_end = name_end_ptr - c_buf
+            sequence_start = name_end + 1
+            sequence_end_ptr = <char *>memchr(c_buf + sequence_start, 10, <size_t>(bufend - sequence_start))
```
marcelm, line 295: I’m not sure, but could bufend - sequence_start ever become negative? I’m just worried about `sequence_start = name_end + 1`. If name_end_ptr points to the last character of the buffer (that is, the previous memchr returned bufend_ptr - 1), then sequence_start will point to the first character *after* the buffer. Or is this not possible?

marcelm (commented):
I’m not finished, but here are a couple of comments already.

src/dnaio/_core.pyx
rhpvorderman: No. Because before that the following check happens: `if name_end_ptr == NULL or name_end_ptr == bufend_ptr:` 
If `name_end_ptr == bufend_ptr` it also means that `name_end == bufend`. In which case we break.
Name_end must therefore be always smaller than bufend. So sequence_start can never point to a character after the buffer.

Having said that, it is good that you mention this, because after checking the qualities newline the check goes thus:
`if qualities_end_ptr == NULL:`. Because it is OK for the FASTQ record to end at bufend. But that means there needs to be a check somewhere to ensure that `next_record_start`, which is `qualities_end+1` does not exceed bufend. That check was removed in my reorganizing of the code. I readded it now.

src/dnaio/_core.pyx
```diff
@@ -258,122 +266,127 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
                 bufstart += 1
                 bufend += 1
                 extra_newline = True
-            else:
-                break
+            elif last_read_position > record_start:  # Incomplete fastq records are present.
+                    if extra_newline:
```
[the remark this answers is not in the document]
rhpvorderman: Yes it can. In case of an incomplete record it will first visit the above if branch. Then it will run trough the parser code, only to find that adding the extra linefeed did not help. It will then trigger this if branch. This is necessary because this will report the incomplete record without the extra newline which was added by dnaio, but was not actually there in the input.
If I remove this check, multiple tests will fail with an off-by-one error because of this. You put the check there for a good reason.

I will add a comment to elucidate, given that the reason for the check is not entirely clear.

src/dnaio/_core.pyx
rhpvorderman: I replaced the `while True` for the inner loop with `while record_start < bufend` to eliminate the possibility of this occurring. I found it more elegant than putting a check at the end of the loop. If there is a `while` statement anyway it is always nice to pair it with a meaningful condition (if the situation calls for it).

src/dnaio/_core.pyx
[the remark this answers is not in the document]
marcelm: Yes, you’re right. I got confused by the original `break` and somehow assumed that all branches exit the loop, but that’s not the case.

src/dnaio/_core.pyx
```diff
@@ -258,122 +266,129 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
-                pos += 1
-            if pos == bufend:
+        while record_start < bufend:
+            ### Check for a complete record (i.e 4 newlines are present)
+            # Use libc memchr, this optimizes looking for characters by
+            # using 64-bit integers. See:
+            # https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=string/memchr.c;hb=HEAD
+            # void *memchr(const void *str, int c, size_t n)
+            name_end_ptr = <char *>memchr(c_buf + record_start, 10, <size_t>(bufend - record_start))
+            if name_end_ptr == NULL or name_end_ptr == bufend_ptr:
```
marcelm, line 293: Can `name_end_ptr == bufend_ptr` ever be True? `bufend_ptr` points to the byte just after the buffer and I wouldn’t expect `memchr` to ever look at this.

src/dnaio/_core.pyx
marcelm: The issue I see is actually a different one, see comment above.

src/dnaio/_core.pyx
rhpvorderman: Indeed. It points at the null byte so it is never true. Well-spotted! I can remove those checks and everything seems fine.

The code will never crash because. 
- If there is no linefeed it returns NULL
- If there is a linefeed at the last character, `name_end` will be at `bufend -1`. It can never be at `bufend` because that is a \00 byte. In that case, if we increase with +1  the next point will be at bufend. Bufend-bufend == 0. When memchr gets a size 0, it will return NULL. So it is safe to do a pointer increase with +1 inbetween memchr calls. But not more than one.

I added these snippets because of `if pos == bufend`, and the code worked fine with no crashes. But thinking about it, they are not necessary.

rhpvorderman: I adopted the code. It also means that testing next_record_start at `sequence_end + 1`. `bufend` is the maximum value and in that case memchr will return NULL.

src/dnaio/_core.pyx
rhpvorderman: Ah. It seems I did not properly think the `bufend_ptr` trough. Sorry.

rhpvorderman:
I made one (minor) organizational change in the code.
`PyUnicode_DecodeLatin1` is now called directly instead of cython's pointer slicing.  This uses pointers and length instead of beginning and start positions. This has the advantage that the end positions do not have to be altered to accomadete for `\r\n` linebreaks. We can adjust the length for that. This also means that `qualities_end` will always be the end of the FASTQ record and `qualities_end +1` is always the start of a new FASTQ record. So this change eliminates the need for the `next_record_start` variable.

I also think that calling the Python C-API in this case looks nicer than cython's pointer slicing and decoding. (Which evaluates to the same code anyway).

rhpvorderman:
Do yo have any estimated time when you can have a look at this? It is a blocker for two other PRs I still want to do:

- Use PyUnicode_DecodeASCII for sequence and qualities. This will incur only a slight performance penalty with the bonus of doing some very basic input sanitization. (Python uses an ASCII mask internally, a 64-bit integer that checks 8 bytes at the same time, so it is quite fast). This PR will be a very small change in the code. The added test code will be much bigger.
- Add the ability to output bytes in a SequenceBytes object. This will be much faster than using strings, and it would benefit my projects downstream, which are simple and use Cython and the C-API anyway, which do interact more easily with bytes objects.

I could start on these PRs using the current code, but that will give very nasty merge conflicts once this is merged.
I realize that "Review my PRs, so I can send you more PRs!" is not very enticing. But I promise that after that dnaio has the functionality I want. (Except for maybe faster paired reading, but that can wait). Anyway, by no means feel pressured to change your schedule. If I have a rough ballpark estimate it is easier for me to focus my mind on other things in the meantime while I shelve this mentally.

marcelm (commented):
You keep sending me easier to review PRs, which is why I didn’t feel bad about ignoring this one until now ... I think I have reviewed everything as thoroughly as I can. Just a couple of minor things and then this can be merged.

src/dnaio/_core.pyx
```diff
@@ -258,122 +266,133 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
-            if pos == bufend:
+            # The maximum value of name_end_ptr = c_buf + bufend - 1.
+            # c_buf + bufend points at the terminating null byte.
+            # c_buf + bufend - 1 is the last character in the buffer. Therefore
+            # we can always safely increase the pointer with +1 to look for the next
+            # newline. Because in case of the end of buffer, name_end will be
+            # bufend - 1. So sequence_start will be bufend. The search
+            # length will be bufend - bufend == 0 and memchr will return NULL.
+            name_end = name_end_ptr - c_buf
+            sequence_start = name_end + 1
```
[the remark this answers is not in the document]
rhpvorderman: Thanks. Comment brevity is not one of my strong points.

src/dnaio/_core.pyx
```diff
@@ -258,122 +266,133 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
+            name_start = record_start + 1  # Skip @
+            second_header_start += 1  # Skip +
+            name_length = name_end - name_start
+            sequence_length = sequence_end - sequence_start
+            second_header_length = second_header_end - second_header_start
+            qualities_length = qualities_end - qualities_start
+
+            # Check for \r\n line-endings and compensate
+            if c_buf[name_end - 1] == b'\r':
+                name_length -= 1
```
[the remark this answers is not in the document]
rhpvorderman: See #26 . The actual parsing of the FASTQ is not where most of the compute power is now. Since the addition of memchr it is actually in creating the string objects.
So I am inclined to keep the "human-readable" version here.
We can benchmark it later indeed, in a separate PR, but I doubt the performance will be noticable.

What I do think matters more here is the actual accessing of the contents of `c_buf`. We have just read `c_buf` from `record_start` to `qualities_end`, but then we jump back again to the positions `name_end`, `sequence_end`, `second_header_end` and `qualities_end`. 
Next we iterate from record_start to sequence_end again when creating the string objects.
So now we traverse the same array of memory three times. Once for finding the newlines, once for checking the line endings and once for the creation of the python objects. I think the newline and line end checking can be combined. But I do wonder how much performance it will add.

rhpvorderman:
Thank you very much for the review @marcelm! Much appreciated.

I have addressed your comments. And I am glad you could not review right away, because it gave me some time to think about the optimal implementation of using `bytes` in this project.

src/dnaio/_core.pyx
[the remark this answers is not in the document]
marcelm: Sure, keep that version, and as I said, I expect the difference to be marginal at best.

Regrouping the statements sounds like an easy thing to do to improve cache locality a little bit. My guess is it won’t help when the reads are short since everything should be in the first-level cache anyway.

marcelm:
This is a great improvement, thanks once more!

PR #16, opened 2021-12-12
Add _needs_stub attribute to allow pip install .

rhpvorderman:
I had some trouble with `pip install .`. This fixes that. I use `pip install .` quite a lot for benchmarks.

EDIT: for python-isal I also had this issue and fixed it this way. Eventually I simply gave up and always required cython installed. With the new-style packages this is quite doable.
The cythonization therefore no longer happens in build_ext, but in the top-level setup.py. This has the advantage of being able to set project-wide compiler directives. (Although that is not relevant for this project with only one cython file).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,10 @@ class BuildExt(build_ext):
             # only sensible thing is to require Cython to be installed.
             from Cython.Build import cythonize
             self.extensions = cythonize(self.extensions)
+            # Setuptools build_meta requires a _needs_stub attribute.
+            # This is required for `pip install .`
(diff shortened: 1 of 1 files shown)


marcelm:
I had the `_needs_stub` issue in WhatsHap last week and also decided to just make Cython a build-time dependency. Everything became just so much nicer, see https://github.com/whatshap/whatshap/pull/344. I made some of the changes in Cutadapt as well, and had just not gotten around to modernizing dnaio. I have now done so in #19 and will therefore close this PR.

Issue #1, opened 2019-06-20
Determining file type from contents rather than name


nickilott:
Hi, I have been trying to run trim_galore on fastq files with naming like *.fastq.1.gz. However there is an error which seems to relate to identifying the file type from the contents i.e.

>>> import dnaio
>>> f = dnaio.open("13714_1#1.fastq.2.gz")

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/<path>/python3.6/site-packages/dnaio/__init__.py", line 83, in open
    file1, fileformat=fileformat, mode=mode, qualities=qualities)
  File "/<path>/python3.6/site-packages/dnaio/__init__.py", line 148, in _open_single
    if file.seekable():
AttributeError: 'PipedGzipReader' object has no attribute 'seekable'

marcelm:
Thanks! This was an oversight in the xopen library, where I recently changed how gzipped files are read. I have fixed this now in https://github.com/marcelm/xopen/commit/aa905e14537ef2b908197f059b044dd7942f4794 . Please update to xopen 0.7.1 to get the fix (should be on PyPI in a couple of minutes) (`pip install --upgrade xopen`).

PR #18, opened 2021-12-13
Decrease buffer size to 128KiB to improve performance when reading from pipes.

rhpvorderman:
Sorry for the PR spam. I found some minor things in the code that each should be considered in isolation.

This is the size used by cat and pigz to read files. This uses less memory. No performance impact on normal files. Faster on reading from pipes.

before:
```
  Time (mean ± σ):      2.075 s ±  0.031 s    [User: 1.953 s, System: 0.122 s]
  Range (min … max):    2.012 s …  2.145 s    20 runs
```
After:
```
  Time (mean ± σ):      2.067 s ±  0.021 s    [User: 1.946 s, System: 0.121 s]
  Range (min … max):    2.030 s …  2.113 s    20 runs
 ```

Gzip files:
Before:
```
Benchmark #1: python dnaio_read.py ~/test/big2.fastq.gz
  Time (mean ± σ):      3.417 s ±  0.045 s    [User: 5.089 s, System: 0.560 s]
  Range (min … max):    3.342 s …  3.487 s    10 runs
```
After
```
python dnaio_read.py ~/test/big2.fastq.gz
  Time (mean ± σ):      3.264 s ±  0.035 s    [User: 5.061 s, System: 0.567 s]
  Range (min … max):    3.219 s …  3.322 s 
```

This is because xopen opens the gzip file in PipedReader mode. With a max pipesize of 1 MB. If the one MB is read entirely while the pipe is not entirely full there might be blocking. This does not happen when reading smaller chunks.

 Also 128 KiB is used by `cat` and `pigz`. These are very good programs.

### DIFF
--- a/src/dnaio/readers.py
+++ b/src/dnaio/readers.py
@@ -121,7 +121,7 @@ class FastqReader(BinaryFileReader, SingleEndReader):
         file: Union[str, BinaryIO],
         *,
         sequence_class=Sequence,
-        buffer_size: int = 1048576,
+        buffer_size: int = 128 * 1024,  # Buffer size used by cat, pigz etc.
         opener=xopen,
         _close_file: Optional[bool] = None,
     ):


marcelm:
I’ll start with this one since it’s an easy one :-)

PR #21, opened 2021-12-15
Use zip_longest for more readable and faster code in the paired end iterator

rhpvorderman:
Sorry for the PR spam. I just suddenly find so much stuff to optimize in this library. And even more is coming, but that depends on still pending PRs #14 and #20. (I want to make writing faster still, by avoiding creating new bytes objects).

This is an easy one to review. I know you are busy!

Ideally we would want to use `zip(it1, it2, strict=True)`, but that is only available from python 3.10 and higher. (3.9 on my Debian 11 machine).

Since try ... except has some expense I tried using `next` differently. It has a default argument. So you can use `next(it1, None)` and it will yield None if the iterator is exhausted. This leads to a cleaner looking code when checking.

I had a look at itertools if there was something in the standard library, and yes there was: zip_longest. Keeps on yielding until the longest iterator is exhausted and yields None for the shortest iterator. 

The code looks cleaner now now all the exhaustion checking is in zip_longest instead of in try except loops. As an added bonus, itertools is fully implemented in C so it is faster as well:

before:
```
  Time (mean ± σ):      4.884 s ±  0.031 s    [User: 4.667 s, System: 0.216 s]
  Range (min … max):    4.826 s …  4.919 s    10 runs
 ```
after
```
  Time (mean ± σ):      4.746 s ±  0.049 s    [User: 4.540 s, System: 0.206 s]
  Range (min … max):    4.691 s …  4.835 s    10 runs
```

It is 2-3% faster. Not bad for such a small change.

### DIFF
--- a/src/dnaio/pairedend.py
+++ b/src/dnaio/pairedend.py
@@ -54,19 +54,19 @@ class TwoFilePairedEndReader(PairedEndReader):
         # but it will take a long time for 3.10 or higher to be available on
         # everyone's machine.
         # Instead use zip_longest from itertools. This yields None if one of
-        # the iterators is exhausted. Checking for truthy values is speedy.
+        # the iterators is exhausted. Checking for None identity is fast.
         # So we can quickly check if the iterator is still yielding.
         # This is faster than implementing a while loop with next calls,
         # which requires expensive function lookups.
         it1, it2 = iter(self.reader1), iter(self.reader2)
         for r1, r2 in itertools.zip_longest(it1, it2):
-            if not r1:
+            if r1 is None:
                 raise FileFormatError(
                     "Reads are improperly paired. There are more reads in "
                     "file 2 than in file 1.",
                     line=None,
                 ) from None
-            if not r2:
+            if r2 is None:
                 raise FileFormatError(
                     "Reads are improperly paired. There are more reads in "
                     "file 1 than in file 2.",
--- a/src/dnaio/pairedend.py
+++ b/src/dnaio/pairedend.py
@@ -1,3 +1,4 @@
+import itertools
 from contextlib import ExitStack
 from os import PathLike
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/dnaio/pairedend.py [hunk 1]
```diff
@@ -47,26 +48,25 @@ def __iter__(self) -> Iterator[Tuple[Sequence, Sequence]]:
...
-                        line=None,
-                    ) from None
-                except StopIteration:
-                    pass
-                break
-            try:
-                r2 = next(it2)
-            except StopIteration:
+        for r1, r2 in itertools.zip_longest(it1, it2):
+            if not r1:
```
marcelm, line 63: I think this needs to be `if r1 is None:`:
```python
>>> from dnaio import Sequence
>>> bool(Sequence("empty", "", ""))
False
```

Can you add a test?

rhpvorderman: Ah, that would have gone horribly wrong indeed. It is now `is None` and a test was added.

src/dnaio/pairedend.py
```diff
@@ -47,26 +48,25 @@ def __iter__(self) -> Iterator[Tuple[Sequence, Sequence]]:
...
-                        "file 2 than in file 1.",
-                        line=None,
-                    ) from None
-                except StopIteration:
-                    pass
-                break
-            try:
-                r2 = next(it2)
-            except StopIteration:
+        for r1, r2 in itertools.zip_longest(it1, it2):
```
marcelm, line 62: You can simplify this further by iterating directly over `self.reader1` and `self.reader2`

rhpvorderman: Indeed, I missed that. Done.

marcelm:
Good improvement, thanks once again! I think the comment at the top of `__iter__` is a bit verbose, but let’s leave it as it is for now.

PR #22, opened 2021-12-15
Even faster name checking (slightly).

rhpvorderman:
This improves the speed of name checking by 10% (45ns to 41ns). It is not much. But it counts given the sheer number of records.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -381,7 +381,7 @@ def record_names_match(header1: str, header2: str):
     """
     Check whether the sequence record ids id1 and id2 are compatible, ignoring a
     suffix of '1', '2' or '3'. This exception allows to check some old
-    paired-end reads that have names ending in '/1' and '/2'. Also, the
+    paired-end reads that have IDs ending in '/1' and '/2'. Also, the
     fastq-dump tool (used for converting SRA files to FASTQ) appends '.1', '.2'
     and sometimes '.3' to paired-end reads if option -I is used.
     """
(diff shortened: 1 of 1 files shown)


src/dnaio/_core.pyx
```diff
@@ -399,21 +400,30 @@ def record_names_match(header1: str, header2: str):
...
-    if header1_ends != header2_ends:
-        return False
+
+    if header1_ends > header2length:
+        return False  # Header2's name is shorter.
+
+    # No need to search whitespace in header2. We can check if there is
+    # whitespace or a 0-byte at the expected position
+    # Use optimized ordering of and statements. Most likely byte is 0. (No
+    # comments in the header). So that will end the comparison immediately.
```
marcelm, line 423: All Illumina paired-end reads have a comment, so I think this isn’t accurate. My suggestion would be to just remove these two comment lines – unless you benchmark it and find a difference between checking for `'\0'` or `' '` first.

src/dnaio/_core.pyx
```diff
@@ -399,21 +400,30 @@ def record_names_match(header1: str, header2: str):
...
 
 
-cdef bint record_name_bytes_match(char *header1chars, char *header2chars):
+cdef bint record_name_bytes_match(char *header1chars, char *header2chars,
+                                  size_t header2length):
     """
-    Check whether the ascii-encoded names match.
+    Check whether the ascii-encoded names match. Only header2length is needed.
     """
     # Only the first part (i.e. the name without the comment) is of interest.
```
marcelm, line 412: When I added the `--rename` option to Cutadapt, I had to decide on a term for "the part of the read header before the first whitespace" and settled on "id" (or read id). I’d like to use these terms consistently in dnaio as well at some point, so you could write "Only the read ID is of interest".

src/dnaio/_core.pyx
rhpvorderman: Not in my illumina test data, but that is still Hiseq 4000 data. Anyway, I removed the comment as you suggested. You are probably correct there is little optimization value whether 1, 2 or 3 characters need to be matched.

src/dnaio/_core.pyx
rhpvorderman: I renamed everything to "id". It makes the function easier to read too.

src/dnaio/_core.pyx
marcelm: You’re right, it depends on the CASAVA version used. I mainly see data produced by CASAVA 1.8+, which [produces read headers with comments](https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/FileFormat_FASTQ-files_swBS.htm).

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
     # find the spaces and tabs easily.
     cdef char * header1chars = <char *>PyUnicode_1BYTE_DATA(header1)
     cdef char * header2chars = <char *>PyUnicode_1BYTE_DATA(header2)
-    return record_name_bytes_match(header1chars, header2chars)
+    cdef size_t header2length = <size_t>PyUnicode_GET_LENGTH(header2)
+    return record_name_bytes_match(header1chars, header2chars, header2length)
```
marcelm, line 404: Perhaps `record_ids_match` would be better now?

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
...
 
 
-cdef bint record_name_bytes_match(char *header1chars, char *header2chars):
+cdef bint record_name_bytes_match(char *header1chars, char *header2chars,
+                                  size_t header2length):
     """
-    Check whether the ascii-encoded names match.
```
marcelm, line 407: ascii -> ASCII

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
...
-    # Only the first part (i.e. the name without the comment) is of interest.
+    # Only the read ID is of interest.
     # Find the first tab or space, if not present, strcspn will return the
     # position of the terminating NULL byte. (I.e. the length).
-    cdef size_t header1_ends = strcspn(header1chars, b' \t')
-    cdef size_t header2_ends = strcspn(header2chars, b' \t')
-    # Quick check if the lengths match
-    if header1_ends != header2_ends:
-        return False
+    cdef size_t id1_ends = strcspn(header1chars, b' \t')
```
marcelm, line 415: The variable naming is a bit inconsistent: `id1_ends` vs. `header1chars` (I’d prefer including the underscore)

marcelm: And perhaps calling the function parameters just `header1` and `header2` is sufficient.

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
...
+    if id1_ends > header2length:
+        return False  # Header2's name is shorter.
+
+    # No need to search whitespace in header2. We can check if there is
+    # whitespace or a 0-byte at the expected position
+    cdef char end = header2chars[id1_ends]
+    if end != b'\000' and end != b' ' and end != b'\t':
+        return False  # Header2's name is longer.
 
     # check if the names end with 1, 2 or 3. (ASCII 49, 50 , 51)
```
marcelm, line 426: Remove the text in parentheses (since the ASCII values themselves don’t appear anymore)

src/dnaio/_core.pyx
marcelm: ... and it should be "IDs", not "names"

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
...
+    cdef size_t header2length = <size_t>PyUnicode_GET_LENGTH(header2)
+    return record_name_bytes_match(header1chars, header2chars, header2length)
 
 
-cdef bint record_name_bytes_match(char *header1chars, char *header2chars):
+cdef bint record_name_bytes_match(char *header1chars, char *header2chars,
+                                  size_t header2length):
     """
-    Check whether the ascii-encoded names match.
+    Check whether the ascii-encoded names match. Only header2length is needed.
```
marcelm, line 410: I find it a little bit suprising that the length of the *second* header is needed. This could use the length of the first header instead.

src/dnaio/_core.pyx
```diff
@@ -399,28 +400,35 @@ def record_names_match(header1: str, header2: str):
...
     # position of the terminating NULL byte. (I.e. the length).
-    cdef size_t header1_ends = strcspn(header1chars, b' \t')
-    cdef size_t header2_ends = strcspn(header2chars, b' \t')
-    # Quick check if the lengths match
-    if header1_ends != header2_ends:
-        return False
+    cdef size_t id1_ends = strcspn(header1chars, b' \t')
+
+    if id1_ends > header2length:
+        return False  # Header2's name is shorter.
```
marcelm, line 418: The comment has become outdated (it should now be "The id of `header2` is shorter"). I try to avoid comments that essentially just re-state what the code does and try to make the code clearer if possible. If you rename `id1_ends` to `id1_length`, then this can become:
```
if header2_length < id1_length:
```
... and then no comment is needed at all.

marcelm (commented):
Sorry for even more review comments. This is essentially just nitpicking about names and comments.

src/dnaio/_core.pyx
rhpvorderman: Indeed. I refactored the code to `strcspn` header2 instead. This makes the API more intuitive. 
I also added a comment why only header1_length is needed.

rhpvorderman:
Thanks for the good suggestions. The code is much more readable now.
There is just one redundant comment, where it explains what `memcmp` does. People who know `memcmp` can see that from the code directly. But I guess not everybody from the Python world knows `memcmp` so I opted to leave the comment in.

marcelm:
Looks good now!

> But I guess not everybody from the Python world knows memcmp so I opted to leave the comment in.

This is totally fine.

PR #27, opened 2022-01-10
Create Sequence objects faster by bypassing cinit.

rhpvorderman:
When creating a Sequence object, what we want:
- Create a Sequence object
- Set its `name`, `sequence` and `qualities` attributes to the appropriate PyObject pointers.

What this PR does: 
- See above

What the original Cython behaviour is.
- Create a tuple with name, sequence, and qualities
- Pass this to the tp_new function of Sequence.
- Inside tp_new, set `name`, `sequence` and `qualities` to None. Increase the refcount of None by 1. Three times.
- Inside tp_new, call cinit with `args` and `kwargs`
- Inside cinit do a thorough checking to ensure that in fact we have passed three arguments.
- Set `name`, `sequence` and `qualities` attributes to the appropriate PyObject pointers.
- Make sure the refcount of None is decreased.

Especially the argument parsing in Python incurs a very heavy overhead. With this simple change, Sequence objects are created the way they should be and much faster. (Saves about 10% on the total parsing time, measured against #15).


marcelm:
I need to understand this better before I feel comfortable with it. Actually, I wonder whether this could be solved in Cython instead. That is, perhaps there is another way to bypass the overhead of `__new__` without resorting to pure C. Or perhaps it’s time for a feature request for Cython.

My Cython generates this Sequence struct, and I wonder whether it’s a problem that this and the definition in the `.h` file don’t quite match:
```
struct __pyx_obj_5dnaio_5_core_Sequence {
  PyObject_HEAD
  struct __pyx_vtabstruct_5dnaio_5_core_Sequence *__pyx_vtab;
  PyObject *name;
  PyObject *sequence;
  PyObject *qualities;
};
```

rhpvorderman:
https://cython.readthedocs.io/en/stable/src/userguide/extension_types.html#public-extension-types

It is possible! This should be doable.

rhpvorderman:
This will require moving Sequence to its own file. So I will wait for this after #23 is implemented. Otherwise there is going to be merge conflicts.

rhpvorderman:
After giving this some consideration and doing some research, I must conclude that this optimisation is not viable. Cython's structs are not stable. Therefore these cannot be reliably initialized externally from cython. The only option to remedy this is to rewrite sequence in C. And that is simply too much effort, for too little gain.

Issue #26, opened 2021-12-23
Profiling of fastq iter method


rhpvorderman:
Hi, I profiled the fastq iter method by using hyperfine and commenting out code. This was done on a branch based on the PR in #15 . 

The script used was simple:
```
#!/usr/bin/env python3

import dnaio
import sys

with dnaio.FastqReader(sys.argv[1]) as records:
    for record in records:
        pass
```

### Profile

When viewing these tables, I took it as a rule of thumb that a WGS sequence file contains 1 billion records. Therefore a time of 1 ns per read corresponds to 1 second when traversing a WGS file.

Area | Time for 5 million reads (milliseconds) | time for one read (nanoseconds, estimate) | Additional remarks
---|--:|--:|---
Startup dnaio and determine `two_headers` | 37 |  |
Finding start and end positions of FASTQ record and basic testing | 250  | 50 | 140ms user 110ms system (!)
Yielding an object and iterating over a for loop | 150 | 30 |
Creating a tuple | 5  | 1 |
Creating a Sequence object | 100  | 20 |
Creating three `bytes` objects | 280 | 56 |
Creating three `str` objects (latin-1) | 580 | 116 |
Creating three `str` objects (name: latin-1, sequence and qualities: ASCII | 640 | 128 |

The following benchmark were done by adapting the read script and check the differences between tuples and sequence objects.
Area | Time for 5 million reads (milliseconds) | time for one read (nanoseconds, estimate) | Additional remarks
---|--:|--:|---
Unpack tuple in for loop | 100 | 20 |  `for name, sequence, qualities in records`
Access individual tuple member by index |  90 | 18 | `record[1]`
Access individual tuple member by global variable index | 130 | 26 | `record[SEQUENCE]`
Passing a tuple as an argument using `*` | 100 | 20 | `my_func(*record)`
Access Sequence member by name | 160 | 32 | `record.sequence`

### Conclusions

- The core parsing loop that checks all the beginnings and ends of name, sequence, second_header and qualities takes only 50 ns per read. More than 40% of which is system time. This means there is less than 30 ns to be saved here. Further optimizations to the parser code are therefore not worth it (such as the quality length shortcut in #14)
- Bytes objects are created more than twice as fast as string objects.
- Decoding into ASCII is 10% slower than latin-1. Not 50%. The comment in `_core.pyx` is outdated. Currently an ASCII mask is used to check for ASCII values. (It is possible that each individual character was checked in older versions of python, explaining the difference).
[...]

PR #30, opened 2022-01-14
Add a mode for reading sequence as bytes data.

rhpvorderman:
Hi, ~I have not added tests yet, (will do that later), but~, here is the code for using a SequenceBytes object which uses bytes instead of strings. This can be initiated simply by using `dnaio.open(my.fastq, mode="rb")`.

I benchmarked the performance with the following scripts:

##### dnaio_read.py
```python
import dnaio
import sys

with dnaio.open(sys.argv[1]) as records:
    for record in records:
        pass
```
##### dnaio_read_and_write.py
```python
import dnaio
import sys
import xopen

with dnaio.open(sys.argv[1], mode="r") as in_records:
    with xopen.xopen(sys.argv[2], mode="wb") as out_file:
        for record in in_records:
            out_file.write(record.fastq_bytes())
```

The binary variants are the same, just with `mode="rb"`.

Performance using default mode:
```
  Time (mean ± σ):      1.154 s ±  0.016 s    [User: 1.036 s, System: 0.118 s]
  Range (min … max):    1.125 s …  1.178 s    10 runs

  Time (mean ± σ):      1.963 s ±  0.022 s    [User: 1.792 s, System: 0.172 s]
  Range (min … max):    1.930 s …  1.994 s    10 runs
```
Very good already. 1150 ms for reading and about 800 ms for writing.

Performance using the new binary mode:
```
Benchmark #1: python dnaio_read_binary.py ~/test/big2.fastq
  Time (mean ± σ):     795.4 ms ±  20.8 ms    [User: 674.6 ms, System: 120.8 ms]
  Range (min … max):   767.8 ms … 833.7 ms    10 runs

Benchmark #1: python dnaio_read_and_write_binary.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):      1.555 s ±  0.020 s    [User: 1.389 s, System: 0.165 s]
  Range (min … max):    1.532 s …  1.594 s    10 runs
```
Approximately 800 ms for reading -> 30% time reduction.
Approximately 750 ms for writing -> slightly faster. (The fastq_bytes method is about 10 ns faster, so that adds up to about 50 ms on 5 million records).

~~Absolutely stellar performance. I did not expect the increase from the writing part, as we use the 1BYTE pointers directly already... Still very welcome :D.~~ Just what I thought, the previous performance listed (write speed 530 ms) was a bug. The sequence and qualities were truncated to name length due to a copy paste error. And the test did not catch that as name sequence and qualiies all have the same length there.

EDIT: Oh I forgot to add: the aded logic does not affect the speed of the default mode. The speed is the same compared to main.


rhpvorderman:
@marcelm, I checked the test suite and all of the necessary tests are covered. It is ready for review.

marcelm:
Sorry for the delay, but this is hard for multiple reasons.

30% faster reading is great, but I wonder whether it’s worth the added complexity, code size and maintenance burden. I’m therefore not super enthusiastic about this change, but you may be able to change my mind.

Here are a couple of downsides. None of these are dealbreakers per se, though:

- I cannot review `_sequence.c` properly as I don’t have enough experience writing pure-C Python extension modules.
- It’s inconsistent that now some parts of the library are in C and some in Cython.
- I still don’t think that using bytes for the sequence is a good user interface. Using bytes as strings just doesn’t work very well in Python 3 (indexing with an integer gives an int, not a character, for example). And even if I found the tradeoff to be acceptable for the `sequence` and `qualities` attributes, storing the `name` as bytes just feels very off.

Do you have an idea where does the speed difference actually come from? I guess you get 10% from the faster object creation you suggested in #27.

I thought switching to `PyUnicode_DecodeLatin1` would have made constructing unicode objects nearly as fast as bytes objects already. Looking at the [definition for _PyUnicode_FromUCS1](https://github.com/python/cpython/blob/d7c68639795a576ff58b6479c8bb34c113df3618/Objects/unicodeobject.c#L2462), I don’t see anything obvious that would  be, except perhaps for `ucs1lib_find_max_char`, which I assume iterates over the string. (Or perhaps just initializing a unicode object with `PyUnicode_New` is expensive.)

I realize you invested a lot of time, and I’m open for further discussion, but I cannot just merge this right away.

rhpvorderman:
I understand. 

>  I cannot review _sequence.c properly as I don’t have enough experience writing pure-C Python extension modules.
> It’s inconsistent that now some parts of the library are in C and some in Cython.

Yes. The sequence code, I can remove that and simply use Cython instead. This will make the performance differences slightly smaller. 

> Do you have an idea where does the speed difference actually come from? I guess you get 10% from the faster object creation you suggested in #27.

10% from the faster object creation, the other 20% because bytesobjects are faster.

> I thought switching to PyUnicode_DecodeLatin1 would have made constructing unicode objects nearly as fast as bytes objects already. Looking at the definition for _PyUnicode_FromUCS1, I don’t see anything obvious that would be, except perhaps for ucs1lib_find_max_char, which I assume iterates over the string. (Or perhaps just initializing a unicode object with PyUnicode_New is expensive.)

No it does not. See #26. In fact it is two times slower. This is because, creating a bytes object only needs a call to memcpy, which does not check individual bytes but copies using word-size (64-bit) chunks. `ucs1lib_find_max_char` probably checks chars individually so that is massively expensive. Also NULL bytes are illegal in strings, so probably there is some check for that as well.

> I still don’t think that using bytes for the sequence is a good user interface. Using bytes as strings just doesn’t work very well in Python 3 (indexing with an integer gives an int, not a character, for example). And even if I found the tradeoff to be acceptable for the sequence and qualities attributes, storing the name as bytes just feels very off.

See umitools-extract. This tool reads a umi from the sequence parts, and then appends it to the name. This is much easier if both are bytes (or strings for that matter). 

Regarding the bytes, these are so much more convenient when dealing with them in C or Cython. Bytes objects support the buffer protocol, which makes access to their internal pointer very easy and convenient. That is not the case for strings. So when doing high performance stuff, it is very useful to have bytes. I want to note that using the SequenceBytes object is completely optional and not the default mode in dnaio. 

I will open a seperate PR with the SequenceBytes in Cython. (Should be much less intense to review). I do want to note that since Sequence and SequenceBytes are virtually the same, it should be quite easy to also have Sequence in C and get the faster object creation benefit there. But that might be something for a later point.

marcelm:
> > I cannot review _sequence.c properly as I don’t have enough experience writing pure-C Python extension modules.
> > It’s inconsistent that now some parts of the library are in C and some in Cython.
> 
> Yes. The sequence code, I can remove that and simply use Cython instead. This will make the performance differences slightly smaller.

Please wait a little bit. I don’t want you to do all this extra work when we then might decide that switching everything to C is the better option.

> > I thought switching to PyUnicode_DecodeLatin1 would have made constructing unicode objects nearly as fast as bytes objects already.
> 
> No it does not. See #26. In fact it is two times slower.

Thanks for the reminder about #26, I admit I had forgotten that issue.

> This is because, creating a bytes object only needs a call to memcpy, which does not check individual bytes but copies using word-size (64-bit) chunks. `ucs1lib_find_max_char` probably checks chars individually so that is massively expensive.

"massively" is a bit of an overstatement, but yes, that is probably responsible. I’ll open a PR trying to remedy this.

> Also NULL bytes are illegal in strings, so probably there is some check for that as well.

Null bytes are fine in Python strings:
```
$ python3 -c 'print(repr("Hello\0there"))'
'Hello\x00there'
```

> > I still don’t think that using bytes for the sequence is a good user interface. Using bytes as strings just doesn’t work very well in Python 3 (indexing with an integer gives an int, not a character, for example). And even if I found the tradeoff to be acceptable for the sequence and qualities attributes, storing the name as bytes just feels very off.
> 
> See umitools-extract. This tool reads a umi from the sequence parts, and then appends it to the name. This is much easier if both are bytes (or strings for that matter).

Ok, having all of the same type is indeed easier in that case. I’ll give you that one.

> Regarding the bytes, these are so much more convenient when dealing with them in C or Cython. Bytes objects support the buffer protocol, which makes access to their internal pointer very easy and convenient. That is not the case for strings. So when doing high performance stuff, it is very useful to have bytes. I want to note that using the SequenceBytes object is completely optional and not the default mode in dnaio.

Ok, that is something I can buy.

> I will open a seperate PR with the SequenceBytes in Cython. (Should be much less intense to review). I do want to note that since Sequence and SequenceBytes are virtually the same, it should be quite easy to also have Sequence in C and get the faster object creation benefit there. But that might be something for a later point.

Yes, I wonder whether that may be the better option. I’m not opposed to pure C modules. I’m just worried I’ll miss things that Cython does for us.

I’m slightly less opposed now, will think more about this later.

rhpvorderman:
> Yes, I wonder whether that may be the better option. I’m not opposed to pure C modules. I’m just worried I’ll miss things that Cython does for us.

Reference counting. But I wrote a test for that. It is a fairly easy object to make in C.  

> I’m slightly less opposed now, will think more about this later.

Thanks! Take your time.

PR #32, opened 2022-01-25
Construct unicode objects with PyUnicode_New

marcelm:
This avoids some of the overhead of calling `PyUnicode_DecodeLatin1` by creating the unicode object with an unitialized buffer and then copying the data into it.

On my test data, this gives a ~9% speedup for the `for r in dnaio.open("r.fastq"): pass` loop (2.42 seconds to 2.20 seconds).

This is marked as draft because PyUnicode_New can return NULL, in which case we need to raise a `MemoryError()` (this is also done incorrectly in some other places in the code).

I have no idea whether this is done correctly. For example, `PyUnicode_New` actually returns a `PyObject*`, but I only got it to work by declaring it as returning `object`.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -1,5 +1,6 @@
 # cython: language_level=3, emit_code_comments=False
 
+from cpython.ref cimport PyObject
 from cpython.bytes cimport PyBytes_FromStringAndSize, PyBytes_AS_STRING, PyBytes_GET_SIZE
 from cpython.unicode cimport PyUnicode_DecodeLatin1
 from libc.string cimport strncmp, memcmp, memcpy, memchr, strcspn
@@ -388,10 +389,12 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
             # Constructing objects with PyUnicode_New and memcpy bypasses some of
             # the checks otherwise done when using PyUnicode_DecodeLatin1 or similar
             name = PyUnicode_New(name_length, 255)
-            memcpy(PyUnicode_1BYTE_DATA(name), c_buf + name_start, name_length)
             sequence = PyUnicode_New(sequence_length, 255)
-            memcpy(PyUnicode_1BYTE_DATA(sequence), c_buf + sequence_start, sequence_length)
             qualities = PyUnicode_New(qualities_length, 255)
+            if <PyObject*>name == NULL or <PyObject*>sequence == NULL or <PyObject*>qualities == NULL:
+                raise MemoryError()
+            memcpy(PyUnicode_1BYTE_DATA(name), c_buf + name_start, name_length)
+            memcpy(PyUnicode_1BYTE_DATA(sequence), c_buf + sequence_start, sequence_length)
             memcpy(PyUnicode_1BYTE_DATA(qualities), c_buf + qualities_start, qualities_length)
 
             if n_records == 0:
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -10,6 +10,7 @@ cdef extern from *:
     unsigned char * PyUnicode_1BYTE_DATA(object o)
     int PyUnicode_KIND(object o)
     int PyUnicode_1BYTE_KIND
+    object PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar)
 from .exceptions import FastqFormatError
 from ._util import shorten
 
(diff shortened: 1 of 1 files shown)


rhpvorderman:
`object` is the alias for `PyObject *` in Cython. Code seems to work well.
I would declare the new objects all in the same block, so you can do this:
```
name = PyUnicode_New(name_length, 255)
sequence = <idem>
qualities = <idem>
if name == NULL or sequence == NULL or qualities == NULL:
    raise MemoryError()
```
The memory error should be extremely rare and it should not matter which object was too much in the end. So this is a fast way to check.

marcelm:
Thanks for the suggestion! (I had to cast to `<PyObject*>`, though.)

> `object` is the alias for `PyObject *` in Cython.

I was confused because there is a a [difference when casting](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#type-casting):
> Casting to `<object>` creates an owned reference. Cython will automatically perform a Py_INCREF and Py_DECREF operation. Casting to `<PyObject *>` creates a borrowed reference, leaving the refcount unchanged.

So I wasn’t sure whether Cython makes any incorrect assumptions about the reference count when the function is declared with a return type of object.

rhpvorderman:
I benchmarked this against the `bytes` stuff, and it is quite good. 1130 ms for `main`, 980 ms for this change and about 920ms for the bytes stuff. So this makes the Unicode objects almost on par with the bytes.

Sorry for my short remarks initially, I had to go somewhere. I will review now.

src/dnaio/_core.pyx
```diff
@@ -10,6 +10,7 @@ cdef extern from *:
     unsigned char * PyUnicode_1BYTE_DATA(object o)
     int PyUnicode_KIND(object o)
     int PyUnicode_1BYTE_KIND
+    object PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar)
```
rhpvorderman, line 13: PyUnicode_New will be in Cython from version 3.0 onwards. (Unfortunately cython still supports python2, so it is not there yet). So we can move this block to the Cython imports when the time comes.

src/dnaio/_core.pyx
```diff
@@ -384,15 +385,16 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
-            # This is because PyUnicode_DecodeLatin1 is an alias for
-            # _PyUnicode_FromUCS1. Which directly copies the bytes into a
-            # string object after some checks. With PyUnicode_DecodeASCII,
-            # there is an extra check whether characters exceed 128.
-            name = PyUnicode_DecodeLatin1(c_buf + name_start, name_length, 'strict')
-            sequence = PyUnicode_DecodeLatin1(c_buf + sequence_start, sequence_length, 'strict')
-            qualities = PyUnicode_DecodeLatin1(c_buf + qualities_start, qualities_length, 'strict')
+            # Constructing objects with PyUnicode_New and memcpy bypasses some of
+            # the checks otherwise done when using PyUnicode_DecodeLatin1 or similar
+            name = PyUnicode_New(name_length, 255)
```
rhpvorderman, line 390: This will change the behaviour as previously PyUnicode_DecodeLatin1 would have found the highest character to be 127. So it would know the object to be of type `PyASCIIObject`. Now it will be another type.
I don't know if that is a very  significant change.

rhpvorderman (approved):
Ready to merge!

My only remark is that we might consider some basic sanity checking in the feature. Sequence and qualities should not have characters outside the ASCII range, and that check is pretty cheap. So we might consider that.

rhpvorderman:
Also, thanks for looking this up. I am writing a faster than pysam BAM parser in my spare time (currently on the [init tree](https://github.com/rhpvorderman/pybam/tree/init)). This comes in useful for the read name. I store it as `bytes` (fastest), but it makes sense to also offer a convenient `str` property that the user can access.

src/dnaio/_core.pyx
```diff
@@ -384,15 +385,16 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
...
+            sequence = PyUnicode_New(sequence_length, 255)
+            qualities = PyUnicode_New(qualities_length, 255)
+            if <PyObject*>name == NULL or <PyObject*>sequence == NULL or <PyObject*>qualities == NULL:
+                raise MemoryError()
+            memcpy(PyUnicode_1BYTE_DATA(name), c_buf + name_start, name_length)
```
rhpvorderman, line 395: Apparently [PyUnicode_DATA](https://docs.python.org/3.11/c-api/unicode.html#c.PyUnicode_DATA) will also work as a `void *`. Maybe it is faster?

src/dnaio/_core.pyx
marcelm: Yes, I realize this changes the KIND. I think the difference is that a `PyASCIIObject` can safely return a pointer to its data when asked for the UTF-8 representation. If it’s just latin-1, the UTF-8 representation needs to be computed. But we never use the UTF-8 representation, right? Then there’s no difference in practice.

src/dnaio/_core.pyx
marcelm: Good to know. I’m also looking forward to Cython 3.

src/dnaio/_core.pyx
marcelm: I cannot find its implementation, but `PyUnicode_1BYTE_DATA` is just a macro, so I don’t think this takes any relevant time at all.

marcelm:
> we might consider some basic sanity checking in the feature. Sequence and qualities should not have characters outside the ASCII range, and that check is pretty cheap. So we might consider that.

Wouldn’t that be the same as just using `PyUnicode_DecodeASCII`?

rhpvorderman:
> > we might consider some basic sanity checking in the feature. Sequence and qualities should not have characters outside the ASCII range, and that check is pretty cheap. So we might consider that.
> 
> Wouldn’t that be the same as just using `PyUnicode_DecodeASCII`?
 
Yes. But as shown here, that call is slow compared to PyUnicode_New and an ascii mask is quite cheap. So maybe we can do it faster... But only if we feel if it's worth it.

src/dnaio/_core.pyx
rhpvorderman: You are right: https://github.com/python/cpython/blob/45f5f52601ebccb195c19cb0a77beaf7f7dfa56a/Include/cpython/unicodeobject.h#L320

rhpvorderman:
> My only remark is that we might consider some basic sanity checking in the feature.

Whoops I meant future. Not feature, sorry.

rhpvorderman:
Can this be merged?

marcelm:
Yes

PR #33, opened 2022-01-25
Add a mode for reading sequence as bytes data. (Without C code reprise)

rhpvorderman:
I removed the C code. This makes the creation of the SequenceBytes objects slower.

```
Benchmark #1: python dnaio_read.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):      1.145 s ±  0.023 s    [User: 1.024 s, System: 0.121 s]
  Range (min … max):    1.127 s …  1.203 s    10 runs

Benchmark #1: python dnaio_read_binary.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):     931.3 ms ±  17.7 ms    [User: 809.0 ms, System: 122.0 ms]
  Range (min … max):   907.3 ms … 957.5 ms    10 runs
```

This saves 200 ms on reading and about 50 ms on writing. It also makes it easier when calling C-API/Cython functions as bytes objects have the buffer protocol.


marcelm:
Yes, I think I am fine with this change now in principle. I’ll add a couple of comments already now and will try to find time for a proper review later this week.

src/dnaio/_core.pyx
```diff
@@ -146,6 +150,27 @@ cdef class Sequence:
         return self.fastq_bytes(two_headers=True)
 
 
+cdef class SequenceBytes(Sequence):
```
marcelm, line 153: The naming is typically from most specific to least specific as in `MemoryError`, `FastqReader`, `EncodingWarning`, `PosixPath` etc. so I would call this a `BytesSequence`.

marcelm, line 153: I don’t think deriving from `Sequence` is the proper thing to do. For example, `qualities_as_bytes` won’t work because it calls `self.qualities.encode()`. `__repr__` is also wrong at the moment. Also, since `Sequence.__cinit__` now gets `object` arguments, the type checking done by Cython is lost. Some of this is fixable, but then the main issue remains that `SequenceBytes` cannot be used wherever a `Sequence` is expected, that is, the Liskov Substitution Principle is violated.

src/dnaio/_core.pyx
```diff
@@ -423,30 +457,42 @@ def fastq_iter(file, sequence_class, Py_ssize_t buffer_size):
             buf[0:bufstart] = buf[record_start:bufend]
 
 
-def record_names_match(header1: str, header2: str):
+def record_names_match(header1: Union[str, bytes], header2: Union[str, bytes]):
```
marcelm, line 460: I haven’t fully made up my mind here, but spontaneously, I’m not very fond of this solution and would probably prefer to have a separate function `record_names_match_bytes`.

It would probably make sense to turn this into a method on `Sequence` and `BytesSequence` (but that needs a separate PR).

src/dnaio/singleend.py
```diff
@@ -19,16 +19,16 @@ def _open_single(
...
 
     path: Optional[str]
     if _is_path(file_or_path):
         path = os.fspath(file_or_path)  # type: ignore
-        file = opener(path, mode + "b")
+        file = opener(path, mode[0] + "b")
         close_file = True
     else:
-        if mode == "r" and not hasattr(file_or_path, "readinto"):
+        if 'r' in mode and not hasattr(file_or_path, "readinto"):
```
marcelm, line 31: Please use double quotes

src/dnaio/_core.pyx
rhpvorderman: Yes, that makes sense, as it is a specific type of sequence. Thanks. I will do that.

rhpvorderman: Agreed. I will do a simple copypaste then with some minor alterations.

Sequence and BytesSequence can share a lot, but Cython makes that very hard to do. It can be implemented more easily directly in the Python C-API. But I will do that in a later PR.

src/dnaio/_core.pyx
rhpvorderman: A `id_matches` method seems like a good idea for a later PR indeed.

rhpvorderman: I implemented a separate name matcher. I think the difference between `record_names_match` and `record_names_match_bytes` shows why I desparately want a BytesSequence object. 

This did however require some extra python code in the paired readers. I made sure the two commits are atomic, so they can easily be converted if you like the previous solution better.

src/dnaio/__init__.py
```diff
@@ -73,7 +74,11 @@ def open(
         objects (as str or as pathlib.Path). Use only file1 if data is single-end.
         If sequences are paired, use also file2.
       mode:
-        Either ``'r'`` for reading, ``'w'`` for writing or ``'a'`` for appending.
+        Either ``'r'`` or ``'rb'`` for reading, ``'w'`` for writing
+        or ``'a'`` for appending.
+
+        With mode ``'r'`` the returned Reader iterates over Sequence objects.
+        With mode ``'rb'`` the returned Reader iterates over SequenceBytes objects.
```
marcelm, line 81:
```suggestion
        With mode ``'rb'`` the returned Reader iterates over BytesSequence objects.
```

src/dnaio/_core.pyi
```diff
@@ -14,7 +14,22 @@ class Sequence:
...
+    def fastq_bytes(self) -> bytes: ...
+    def fastq_bytes_two_headers(self) -> bytes: ...
+
 def paired_fastq_heads(buf1: Union[bytes,bytearray], buf2: Union[bytes,bytearray], end1: int, end2: int) -> Tuple[int, int]: ...
 # TODO Sequence should be sequence_class, first yielded value is a bool
-def fastq_iter(file: BinaryIO, sequence_class, buffer_size: int) -> Iterator[Sequence]: ...
+T = TypeVar("T")
+def fastq_iter(file: BinaryIO, sequence_class: Type[T], buffer_size: int = 128 * 1014) -> Iterator[T]: ...
 def record_names_match(header1: str, header2: str) -> bool: ...
+def record_names_match_bytes(header1: bytes, header2: bytes) -> bool: ...
```
marcelm, line 35: A newline is missing at the end of the file

src/dnaio/_core.pyi
```diff
@@ -14,7 +14,22 @@ class Sequence:
...
+    def __len__(self) -> int: ...
+    def __richcmp__(self, other: BytesSequence, op: int) -> bool: ...
+    def fastq_bytes(self) -> bytes: ...
+    def fastq_bytes_two_headers(self) -> bytes: ...
+
 def paired_fastq_heads(buf1: Union[bytes,bytearray], buf2: Union[bytes,bytearray], end1: int, end2: int) -> Tuple[int, int]: ...
 # TODO Sequence should be sequence_class, first yielded value is a bool
-def fastq_iter(file: BinaryIO, sequence_class, buffer_size: int) -> Iterator[Sequence]: ...
+T = TypeVar("T")
+def fastq_iter(file: BinaryIO, sequence_class: Type[T], buffer_size: int = 128 * 1014) -> Iterator[T]: ...
```
marcelm, line 33:
```suggestion
def fastq_iter(file: BinaryIO, sequence_class: Type[T], buffer_size: int = ...) -> Iterator[T]: ...
```

src/dnaio/_core.pyx
```diff
@@ -1,16 +1,20 @@
 # cython: language_level=3, emit_code_comments=False
 
+from cpython.bytes cimport PyBytes_FromStringAndSize, PyBytes_AS_STRING, PyBytes_Check, PyBytes_GET_SIZE
+from cpython.unicode cimport PyUnicode_DecodeLatin1, PyUnicode_Check
```
marcelm, line 4: There’s already a `from cpython.unicode cimport ...` line below, can you please merge them?

tests/test_internal.py
```diff
@@ -44,6 +53,44 @@ def test_fastq_bytes_two_headers(self):
...
+        sequence = b"ACGT"
+        qualities = b"===="
+        name_ref = sys.getrefcount(name)
+        seq_ref = sys.getrefcount(sequence)
+        qual_ref = sys.getrefcount(qualities)
+        seqbytes = BytesSequence(name, sequence, qualities)
+        assert sys.getrefcount(name) == name_ref + 1
+        assert sys.getrefcount(sequence) == seq_ref + 1
+        assert sys.getrefcount(qualities) == qual_ref + 1
+        del(seqbytes)
```
marcelm, line 88:
```suggestion
        del seqbytes
```

marcelm (commented):
Looks good! Only a couple of nitpicks remaining. flake8 is also complaining about unused imports.

rhpvorderman:
All done! Thanks for the review.

marcelm:
Alright, let’s do it!

PR #34, opened 2022-01-25
Fix parsing of FASTA files that contain a comment and no reads

marcelm:
See https://github.com/marcelm/cutadapt/issues/591.

The FASTA parser allows comment lines in FASTA (starting with `#`), but the `_fasta_head` function does not take this into account.

This also makes some error messages somewhat more informative.


rhpvorderman:
Looks good to me!

PR #37, opened 2022-02-04
Implement Sequence and BytesSequence in C.

rhpvorderman:
This shaves of another 100 milliseconds when reading 5 million records. A 10+% improvement. Also it allows more reuse of methods between Sequence and BytesSequence.

I noticed there were no tests for SequenceRecord length checks and for SequenceRecord sliciing.
These are fairly easy to implement, but I wonder if it is necessary if there are no tests? I guess the slicing comes in pretty useful for cutadapt. 

Given the mild speed  increase, this is not an important PR. It is a nice exercise in using the C-API. Getting rid of Cython's overhead is nice too, athough it comes at a cost of verboseness of the code.


rhpvorderman:
We have been in contact about this. This code adds quite a maintenance burden by adding another language and that is simply not worth the speed increase.

PR #41, opened 2022-02-08
Implement ASCII check

### LINKED ISSUE #40: Implement ASCII check
Currently I am not happy with the way we are creating strings, setting maxchar at 255. For FASTQ only ASCII characters are valid, so I feel that we should perform some form of check. The same as would happen if a user would use `open("my.fastq", "rt", encoding="ascii")`. I have several reasons for this:

- Only ASCII is valid in FASTQ
- ASCII strings have the smallest and simplest representation inside the Python C-API. This helps memory use and speed.
- I want to implement stricter checks on the SequenceRecord object. If we set strict ASCII checks on the `__init__` method and the getters and setters, then we can be 100% sure that a SequenceRecord object never contains anything but ASCII strings. That means we can drop a lot of the checks in other methods such as `fastq_bytes` helping speed. Currently this is not possible because the strings we provide are maxchar 255, so that would be incompatible. We can of course make the SequenceRecord object check for maxchar 255, but that does not make sense.

Back when https://github.com/marcelm/dnaio/pull/32 was merged, I noted that we may be able to do it faster than PyUnicode_DecodeASCII. Today I thought of a way to do so.

Checking for ASCII is done with an ASCII mask usually, as ASCII is a 7-bits encoding. The most significant 8th bit is therefore always 0. A mask can be constructed using bytes with the exactvalue of 128 (`0x80` in hex). Since we have 64-bit computers we can test 8 bytes at the time with mask: `0x8080808080808080`. This is what happens in CPython as well when PyUnicode_DecodeASCII is used.

But a 8-byte mask can not always be applied. If we have a read name of 29 characters:
- 29 / 8 = 3
- 29 % 8 = 5
So we perform 3 8-byte checks and 5 1-byte checks for 8 checks  in total.

The numbers for a 150-bp illumina record with a 29-byte name:
- 8 checks for the name
- 150 / 8 = 18, 150 % 8 = 6: 24 checks for sequence
- Same as above: 24 checks for qualities.

A total of 56 checks.

However since we have the entire record in memory we can check the entire record for ASCII. In this case this is 335 characters.
- 335 /8 = 41, 335 % 8 = 7.  

A total of 48 checks.

Not to mention the easier logic if we only apply one ASCII check once for an entire record rather than each SequenceRecord member individually. 

The ASCII check code can be easily programmed in a separate C header file and imported in Cython. It should have minimal impact on performance.

rhpvorderman:
fixes #40 

Couldn't resist to implement it right away.

This implements the ASCII check highlighted in #40. The performance figures:

Current main:
```
  Time (mean ± σ):      1.020 s ±  0.013 s    [User: 909.9 ms, System: 109.6 ms]
  Range (min … max):    1.002 s …  1.043 s    10 runs
```
Using PyUnicode_DecodeASCII
```
  Time (mean ± σ):      1.233 s ±  0.012 s    [User: 1.126 s, System: 0.107 s]
  Range (min … max):    1.220 s …  1.252 s    10 runs
```

~This PR~ EDIT: Outdated. The check is now run at virtually no cost, see below.
```
...
```

~~Running the plain ASCII check takes 80 ms for 5 million records. This is the same difference as I noted between PyUnicode_DecodeLatin1 and PyUnicode_DecodeASCII. Given the small cost, I think this change is worth it.~~


rhpvorderman:
I checked how memchr and CPython handle using 8-bit masks. They align the memory address first. So I did that too. The check is now effectively free:
Current main:
```
  Time (mean ± σ):      1.002 s ±  0.021 s    [User: 883.6 ms, System: 116.7 ms]
  Range (min … max):    0.976 s …  1.037 s    10 runs
```

This PR
```
  Time (mean ± σ):      1.018 s ±  0.031 s    [User: 890.8 ms, System: 126.9 ms]
  Range (min … max):    0.968 s …  1.060 s    10 runs
```

That is not a significant difference, so there is virtually no cost to adding this check.

I have also been looking at the PyUnicode_DecodeASCII implementation. https://github.com/python/cpython/blob/3.9/Objects/unicodeobject.c#L7243

I have no clue why this is so much slower. It checks the length. If it is 0 it returns an empty string reference (same object in python).  If it is 1 it returns a reference to a single character pyobject. If not it creates a new string with PyUnicode_New. Then it calls the `ascii_decode` function which checks for ASCII characters and copies at the same time. It also aligns properly. This is exactly the same as what we do now, except that we do not check for empty and 1-character strings.

The only conclusion that I can come too is that this ascii_decode call is fairly expensive because:

a. it is not in-line (separate function call)
b. By combining the ascii check and copying, it does not allow specific optimisations that memcpy might have.

rhpvorderman:
Hmm, I can't reproduce the 'no cost' of this. There are some compiler shennanigans going on. It seems like dnaio can be 80 ms slower just randomly if code is rearranged while semantically nothing changes. 

There seems to be a cost to this ASCII check (in the 6-24 nanoseconds per 335 bytes range).  I find it hard to determine the actual cost, and therefore also hard to determine if it is worth it.

PR #42, opened 2022-02-09
Implement "is_mate" method on Sequence and BytesSequence

### LINKED ISSUE #35: Add Sequence and BytesSequence method for paired-end id comparison

rhpvorderman:
fixes #35  

Sorry for the PR bombardment. (Again...). Anyway this PR was very easy to make (and hopefully review). This adds an is_mate method to the Sequence and BytesSequence types. This method works with self and other, where other is another object of the same type. 

I see in the Cython generated code that Cython handles this properly: A typecheck on other is performed and Cython knows self and other are structs with a name member. Also Cython properly uses `METH_O` so no argument tuple and dict unpacking is performed. 

The mate check is performed in pairedend.py now with `r1.is_mate(r2)` instead of `record_names_match(r1.name, r2.name)`. 

This saves:
- Two PyObject_GetAttribute calls. (`r1.name` and `r2.name`). These are taken directly from the struct.
- Argument tuple and dict unpacking, as there is only one argument.
- Some checks. The Sequence and BytesSequence types guarantee their members are unicode or bytes respectively. 

This causes a significant speedup of paired-end reading:
Before (Current main):
```
  Time (mean ± σ):      2.768 s ±  0.021 s    [User: 2.560 s, System: 0.208 s]
  Range (min … max):    2.742 s …  2.800 s    10 runs
 
Benchmark #1: python dnaio_read_paired_binary.py ~/test/big2.fastq
  Time (mean ± σ):      2.561 s ±  0.027 s    [User: 2.366 s, System: 0.195 s]
  Range (min … max):    2.511 s …  2.600 s    10 runs
```

After:
```
  Time (mean ± σ):      2.556 s ±  0.016 s    [User: 2.346 s, System: 0.209 s]
  Range (min … max):    2.529 s …  2.584 s    10 runs
 
Benchmark #1: python dnaio_read_paired_binary.py ~/test/big2.fastq
  Time (mean ± σ):      2.314 s ±  0.027 s    [User: 2.117 s, System: 0.196 s]
  Range (min … max):    2.282 s …  2.369 s    10 runs
```

Together with #39 this reduces the overhead of paired-end reading significantly.


rhpvorderman:
@marcelm. This PR is very tiny, but take your time. Paired-end reading still has some overhead, but at least it is at an acceptable level now (20%ish).

src/dnaio/_core.pyx
```diff
@@ -148,6 +148,24 @@ cdef class Sequence:
...
+        cdef:
+            char * header1_chars = NULL
+            char * header2_chars = NULL
+            size_t header1_length
+        # No need to check if type is unicode as it is guaranteed by the type.
+        if PyUnicode_KIND(self.name) == PyUnicode_1BYTE_KIND:
+            header1_chars = <char *>PyUnicode_1BYTE_DATA(self.name)
+            header1_length = <size_t> PyUnicode_GET_LENGTH(self.name)
+        else:
+            raise ValueError(f"name should be ASCII-only. Got {self.name:R}.")
```
marcelm, line 161: I think this line isn’t covered by tests. Can you add one? I think the `:R` won’t work (or will it?). Did you mean `!r` instead?

src/dnaio/_core.pyx
```diff
@@ -148,6 +148,24 @@ cdef class Sequence:
         """
         return self.fastq_bytes(two_headers=True)
 
+    def is_mate(self, Sequence other):
```
marcelm, line 151: This should get a docstring considering we want to make API documentation available soon.

src/dnaio/_core.pyx
```diff
@@ -148,6 +148,24 @@ cdef class Sequence:
...
+        if PyUnicode_KIND(self.name) == PyUnicode_1BYTE_KIND:
+            header1_chars = <char *>PyUnicode_1BYTE_DATA(self.name)
+            header1_length = <size_t> PyUnicode_GET_LENGTH(self.name)
+        else:
+            raise ValueError(f"name should be ASCII-only. Got {self.name:R}.")
+
+        if PyUnicode_KIND(other.name) == PyUnicode_1BYTE_KIND:
+            header2_chars = <char *>PyUnicode_1BYTE_DATA(other.name)
+        else:
+            raise ValueError(f"name should be ASCII-only. Got {other.name:R}.")
```
marcelm, line 166: Same here

src/dnaio/_core.pyx
rhpvorderman: It does not, you are right. I meant !r. (For float formatting `:` is used so Python has me confused now.)

rhpvorderman:
@marcelm done!
The error messages are slightly hypocrite, as we allow basically any latin-1 encoded character as well, not just ASCII. We have to though, as the Sequence objects we produce all have a maxchar value of 255 so all our objects would fail this check if we would be more stringent. 
See #40

marcelm:
Looks good now!

PR #45, opened 2022-02-15
Simplify FASTQWriter

rhpvorderman:
This simplifies the write method. This is slightly *slower*

Before: 
```
  Time (mean ± σ):      1.957 s ±  0.025 s    [User: 1.849 s, System: 0.108 s]
  Range (min … max):    1.916 s …  1.986 s    10 runs
```

After:
```
  Time (mean ± σ):      2.056 s ±  0.057 s    [User: 1.939 s, System: 0.117 s]
  Range (min … max):    2.010 s …  2.174 s    10 runs
```

However in terms of throughput it is not that big of a difference, and there is always the option of using xopen, and using the resulting file and `fastq_bytes` directly:

```
  Time (mean ± σ):      1.566 s ±  0.026 s    [User: 1.455 s, System: 0.110 s]
  Range (min … max):    1.521 s …  1.609 s    10 runs
```

This is much faster than FastqWriter, and it is simply impossible for FastqWriter to reach those speeds anyway. Simplification increases its overhead from 400ms to 500ms, but makes the code much easier to read.


marcelm:
Simplification at the cost of speed hasn’t really been the aim of your previous PRs, why now? Cutadapt supports writing FASTA and FASTQ files, which is why it doesn’t use `fastq_bytes` directly and instead goes through the `FastqWriter` (or a `FastaWriter`).

rhpvorderman:
> Simplification at the cost of speed hasn’t really been the aim of your previous PRs, why now? 

I figured that since FastqWriter is slower than using xopen directly anyway, we might just as well make it simple. But it is only 20 lines of code, so I guess this PR is redundant. I will close this and I apologize for bringing this up.

marcelm:
No worries. I did not have it on the radar that FastqWriter may be slow, so that is good to know.

It’s a bit surprising to me that the overhead is so relatively large since it’s "just" an extra function call and two attribute lookups (-ish). I tried just now to move the `writers.py` into a `writers.pyx` module and that didn’t help too much, so I’ll leave that be for the moment.

rhpvorderman:
> It’s a bit surprising to me that the overhead is so relatively large since it’s "just" an extra function call and two attribute lookups (-ish). I tried just now to move the writers.py into a writers.pyx module and that didn’t help too much, so I’ll leave that be for the moment.

Well, this is tested on exactly 5 million recrods, so it is only 80 ns per record. An attribute lookup is roughly 20 ns and a function call is roughly 40 ns, so two attribute lookups and the extra function indeed add up to that amount.

So that's why I think that FastqWriter cannot be improved much. Sure you can eliminate one of the attribute lookups maybe by aliasing `self._file.write` to `self._write`, but that is about it. The wrapping cost is already very low at 80 ns. The problem is that FASTQ Illumina records are extremely small and therefore very numerous.

marcelm:
> Sure you can eliminate one of the attribute lookups maybe by aliasing self._file.write to self._write, but that is about it.

Yepp, that was the first thing I tried an it indeed didn’t help much.

rhpvorderman:
No worries, pysam can't write FASTQ anyway (I don't see it in the docs) and dnaio is faster than anything else in the python world. 
I benchmarked pysam reading too and ... well it is offtopic, but dnaio is at least 4 times faster. Also gzipped reading is about 3 times faster. So we really should write a paper.

PR #46, opened 2022-02-15
Refactor two header determination

### LINKED ISSUE #38: Two_headers should be a SequenceRecord attribute
Currently dnaio uses a slightly unusual architecure where the first value of fastq_iter is a boolean, not a SequenceRecord. This determines whether all coming fastq headers are printed with two headers. FastqWriter has a rather quirky implementation to determine its write method.

I think this can be best solved by having a boolean attribute to each sequencerecord. This can be set instantly without branching (no if statement). We can then add a fastq_bytes_as_input method, which will print one or two headers based on the boolean attribute. The fastq_bytes_as_input method can then be used by the FastqWriter class.

This will be fairly trivial to implement once the C-code PR is merged.

rhpvorderman:
FastqIter now returns a boolean on first iteration and a sequence record afterwards. This violates the principle of least surprise. 
Two headers is now determined by peeking the file. I did not implement a full-fledged FASTQ parsing algorithm there as any errors in the FASTQ are likely revealed by FastqIter with an appropriate error.

Because of this change FastqReader._iter can now be properly typed and FastqIter can be safely and intuitively be used by other projects.

EDIT: I also found a small mistake in `_core.pyi` that I have corrected here, but I could do that separately if you do not wish to merge this.

fixes  #38


rhpvorderman:
Hmm. This will not be able to correctly determine two headers for long reads on a non-seekable stream. (i.e. when using xopen's pipe methods). The buffer on io.BufferedReader is simply too small. 

I see no easy way to work around this, other than keeping the already existing code.

PR #48, opened 2022-02-16
Add stringent ASCII-checks to Sequence object

### LINKED ISSUE #40: Implement ASCII check

rhpvorderman:
Fixes #40 

Only ASCII strings are valid in FASTQ. This PR:
+ Makes the name, sequence and qualities attributes in Sequence internal hidden _name, _sequence, _qualities.
+ Implements custom getters and setters that do check ASCII for Sequence objects
+ Implements ASCII checks in Sequence `__init__` method.
+ Removes all 1BYTE_KIND and similar checks in all of Sequence's methods. Because of the above points, these are guaranteed.
+ Adds an ASCII-check that checks the entire record so that ASCII strings can be safely created for it in `FastqIter.__next__`. This ASCII check is implemented with a C-function.
+ Removes the str typing in `__cinit__`. This removes any PyUnicode_Checks that cython does. We already ensure the entered objects are valid Python Unicode ASCII strings, so there is no need for these checks.

The speed impact is minimal:
Before:
```
  Time (mean ± σ):     912.6 ms ±  16.3 ms    [User: 797.2 ms, System: 115.3 ms]
  Range (min … max):   894.1 ms … 947.5 ms    10 runs

  Time (mean ± σ):      2.249 s ±  0.017 s    [User: 2.039 s, System: 0.210 s]
  Range (min … max):    2.225 s …  2.277 s    10 runs

  Time (mean ± σ):      1.542 s ±  0.027 s    [User: 1.432 s, System: 0.110 s]
  Range (min … max):    1.504 s …  1.579 s    10 runs
```

After
```
  Time (mean ± σ):     945.9 ms ±  12.5 ms    [User: 828.9 ms, System: 116.9 ms]
  Range (min … max):   930.8 ms … 966.4 ms    10 runs

  Time (mean ± σ):      2.315 s ±  0.015 s    [User: 2.111 s, System: 0.203 s]
  Range (min … max):    2.291 s …  2.328 s    10 runs

  Time (mean ± σ):      1.600 s ±  0.020 s    [User: 1.473 s, System: 0.126 s]
  Range (min … max):    1.562 s …  1.633 s    10 r
```

The advantage of this PR is that the user can **never** create invalid FASTQ or FASTA records anymore without raising an error.

EDIT: Apparently Cython is not very strict and allows setting `None` for any attribute. So `dnaio.Sequence(None, None, None)` would not raise any errors. That is rectified now. Only qualities may be None.


marcelm:
While trying to figure out what the performance impact of this actually is, I got derailed and set up airspeed velocity (asv) for dnaio. Here are the benchmark results so far: https://marcelm.github.io/dnaio-asv/. You should have received an invitation to the repository in case you want to contribute a benchmark.

asv can also be used to compare two commits, and for some reason, I get the result that this PR slows down parsing by 10% (presumably just like #41). Here is the output (showing only results for Python 3.10):
```
       before           after         ratio
     [faba8111]       [65885459]
     <main>           <rhpvorderman-asciicheck2>
+        993±20ms       1.16±0.04s     1.17  benchmarks.EmptyRecords.time_parse_empty_records [.../conda-py3.10]
+         397±9ms         461±20ms     1.16  benchmarks.time_parse_single('paired300bp_1') [.../conda-py3.10]
+      1.82±0.02s       2.11±0.02s     1.16  benchmarks.time_parse_single('single43bp') [.../conda-py3.10]
+         436±6ms         504±10ms     1.16  benchmarks.time_parse_single('paired150bp_1') [.../conda-py3.10]
+        970±30ms       1.09±0.02s     1.12  benchmarks.time_parse_paired('paired300bp') [.../conda-py3.10]

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.
```

I *actually* just wanted to check whether accessing the attributes of the Sequence object is now slower, but unlike in regular Python, that’s not the case. I learned that
- a descriptor implemented in C is very fast
- an attribute declared as `cdef public`/`cdef readonly` in Cython is implemented as a descriptor anyway, so it makes no difference at all (the `Py_tp_getset` slot is filled the same way before and after this PR).

I’ll re-run my benchmarks and maybe a different one. I’m not against some slowdown for the safety this gives, but then it’s probably easier to just use the existing `PyUnicode_...` function for the ASCII check.

marcelm:
Note that the asv page only shows commits that are on main, not those in this PR.

rhpvorderman:
> While trying to figure out what the performance impact of this actually is, I got derailed and set up airspeed velocity (asv) for dnaio. Here are the benchmark results so far: https://marcelm.github.io/dnaio-asv/. You should have received an invitation to the repository in case you want to contribute a benchmark.

Cool! I will check it out.

> I’ll re-run my benchmarks and maybe a different one. I’m not against some slowdown for the safety this gives, but then it’s probably easier to just use the existing PyUnicode_... function for the ASCII check.

On my machine that is slower than our current method. 
Main:
```
Benchmark #1: python dnaio_read.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):     916.1 ms ±  17.2 ms    [User: 801.2 ms, System: 114.8 ms]
  Range (min … max):   889.0 ms … 942.3 ms    10 runs
```
This branch:
```
Benchmark #1: python dnaio_read.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):      1.056 s ±  0.015 s    [User: 933.4 ms, System: 122.7 ms]
  Range (min … max):    1.029 s …  1.082 s    10 runs
```
Using PyUnicode_DecodeASCII
```
Benchmark #1: python dnaio_read.py ~/test/big2.fastq /dev/null
  Time (mean ± σ):      1.301 s ±  0.057 s    [User: 1.170 s, System: 0.126 s]
  Range (min … max):    1.232 s …  1.417 s    10 runs
```

rhpvorderman:
> an attribute declared as cdef public/cdef readonly in Cython is implemented as a descriptor anyway, so it makes no difference at all (the Py_tp_getset slot is filled the same way before and after this PR)

Indeed. The getter is just increasing a reference count and returning an existing pointer, so that is as fast as it can be. The setter however, should be a little slower because it has more checks. But that is okay, since most of these checks are C macros (fast), and FASTQ files are typically read and not modified extensively. 

It is a bit unfortunate that Cython implements the setters so nonsensically. Allowing None to be set to every attribute. Very annoying. I had to write tests for that to explicitly make sure I had implemented it correctly.

rhpvorderman:
I won't do this now, but if we get crazy later, the ascii check can be performed with vectorized instruction using x86 intrinsics. That way 16, 32 or 64 characters can be checked at the same time. It seems like a nice contained use case to use these functions.

marcelm:
> the ascii check can be performed with vectorized instruction using x86 intrinsics

If you do that, you should submit that as a PR to CPython ...

rhpvorderman:
> If you do that, you should submit that as a PR to CPython ...

I already tested  it. There is no advantage over using 64-bit integers and 8-byte masks. The instruction that enables the test is [`_mm_testz_si128`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_testz_si128&ig_expand=7252). But that function internally calculates two values. ZF and CF. So it does two operations where a bit mask does one. I could not measure a difference. I also tried `_mm_testz_si256` but also no difference.

On top of that, this needs special compiler instructions `-msse4.1` for `_mm_testz_si128`. That is OK, there are no x86 cpus that do not support that anymore (2006 or later). But for `_mm_testz_si256` you need `-mavx` and that is quite a bit more modern. 
Plus this won't work for ARM builds and it makes the code more complex, and I simply could not get it to be faster. So it is not worth it.

I did measure the difference between applying an 8-byte mask and a 1-byte mask everywhere and that difference was quite noticable. The parse time was almost doubled using the naive implementation.

marcelm:
So after redoing some of the benchmarks, the results from https://github.com/marcelm/dnaio/pull/48#issuecomment-1046738257 still stand: The slowdown is about 15% on my machine. (And PyUnicode_DecodeASCII is slower.)

So just as a final confirmation before hitting merge: Do you agree this is worth it?

rhpvorderman:
Definitely. Outputting ASCII strings is exactly what the user should expect. If they want unsafe and speedy, the binary mode is available now.
Also the current method of doing things is still faster than v0.7.1 and includes an ASCII-check now. (PyUnicode_DecodeLatin1 is also slower...) .

PR #49, opened 2022-02-20
Implement buffer protocol and use memchr for paired fastq heads

### LINKED ISSUE #24: Possible improvements for paired_fastq_heads
I saw this comment in `_core.pyx`:
```
# It would be nice to be able to have the first parameter be an
# unsigned char[:] (memory view), but this fails with a BufferError
# when a bytes object is passed in.
# See <https://stackoverflow.com/questions/28203670/>
```

This is possible by using the buffer protocol. For a basic implementation see [here](https://github.com/pycompression/python-isal/blob/9fd81d78d4950377298c292576e61a770a43d885/src/isal/isal_zlib.pyx#L145). This acquires the buffer from a python object, and converts the `void *` to `unsigned char *`. 

I also think paired_fastq_heads would benefit from using `memchr` to search the newlines.

rhpvorderman:
Fixes #24 

Given the experience already gathered with memchr and the buffer protocol this was quite easy to implement. It should make things faster.

Also memoryviews can be used now.

EDIT: And it also reduces the generated Cython code from 500K to 416K. :tada: !


src/dnaio/_core.pyx
```diff
@@ -298,41 +299,39 @@ cdef bytes create_fastq_record(char * name, char * sequence, char * qualities,
...
-# when a bytes object is passed in.
-# See <https://stackoverflow.com/questions/28203670/>
 
-ctypedef fused bytes_or_bytearray:
-    bytes
-    bytearray
-
-
-def paired_fastq_heads(bytes_or_bytearray buf1, bytes_or_bytearray buf2, Py_ssize_t end1, Py_ssize_t end2):
+def paired_fastq_heads(buf1, buf2, Py_ssize_t end1, Py_ssize_t end2):
```
marcelm, line 303: If you see my answer on https://stackoverflow.com/questions/28203670/, it is actually possible now to declare the parameters to be of type `const unsigned char[:]`. It would be nicer to let Cython handle getting the memory view.

src/dnaio/_core.pyx
```diff
@@ -298,41 +299,39 @@ cdef bytes create_fastq_record(char * name, char * sequence, char * qualities,
...
+    cdef Py_buffer data1_buffer
+    cdef Py_buffer data2_buffer
+    PyObject_GetBuffer(buf1, &data1_buffer, PyBUF_SIMPLE)
+    PyObject_GetBuffer(buf2, &data2_buffer, PyBUF_SIMPLE)
+
     cdef:
-        Py_ssize_t pos1 = 0, pos2 = 0
         Py_ssize_t linebreaks = 0
-        unsigned char* data1 = buf1
-        unsigned char* data2 = buf2
```
marcelm, line 323: With the parameter changed to `const unsigned char[:]`, this needs to be changed to `unsigned char* data1 = &buf1[0]`.

marcelm (commented):
Since I wrote that "it would be nice" comment about the buffer protocol, Cython has added support for it. I would prefer using Cython’s method for this as it should result in shorter code.

rhpvorderman:
Done! Cython bloats the generate code to 784K now however... What on earth are they doing?! (Gernerating a lot of custom Cython MemoryView implementations it seems). Oh well, it won't matter for performance probably.

marcelm:
> Cython bloats the generate code to 784K now however...

I opened #50.

src/dnaio/_core.pyx
marcelm: Ok, sorry, but by coincidence, there was a discussion on the Cython mailing list about this pattern just today. The problem is that `&buf1[0]` is bounds checked and actually raises an IndexError if `buf1` happens to be an empty string. I think what should work is this:
```
unsigned char* const data1 = buf1
```
If it doesn’t, feel free to revert to the version that explicitly uses the C API.

marcelm: Oh and I think you can also remove the `from cpython.buffer` cimports.

marcelm: Ok, forget it. The weird `const` thing above actually leads to a segfault. Please just revert to [76368ca](https://github.com/marcelm/dnaio/pull/49/commits/76368caab25aaf55cdbfe1765a9dd3c729c885d1).

rhpvorderman: Thanks! I was not happy with this code change for this precise reason. The `cont unsigned char[:]`, just looks weird. Also it does not support `const char[:]` for some reason???

 I am okay with doing Python in Cython, I am okay with doing CPython-API in Cython, but these weird Cython-only constructs always feel off to me. Plus, Cython is not as well maintained as the Python C-API, it is mostly just Stefan Behnel. He does an excellent job, but he can't keep up with the PRs. So I always feel safer not doing Cython-specific calls. It also makes it easier to port code to the C-API or back to pure Python if necessary.

rhpvorderman: The branch was reverted.

marcelm:
Ok, let’s do this!

PR #51, opened 2022-02-23
Do not ship C file in source distribution

### LINKED ISSUE #50: Stop shipping Cython-generated .c files
Cython is now a build-time requirement, so we don’t need to include the generated `.c` files in the sdist anymore.

rhpvorderman:
fixes #50 

This removes the `cythonize` command from setup.py. Cythonize gives more control over how to cythonize extensions, but this is already covered in the `_core.pyx` file itself as it has the compiler_directives set there.

The disadvantage of `cythonize` is that it converts a cython extension into a distutils.Extension. In other words a normal C-extension. Therefore setuptools thinks that the C-source should be included. 

Setuptools.Extension however does recognize `.pyx` files and will convert them to C files itself. So there is no need for cythonize. By this way setuptools realizes that the source file is not the C file but the `.pyx` file and will therefore only include the pyx file.

I also struck the C file from the manifest. Both the Setuptools.Extension and the MANIFEST.in deletion where needed to remove the C file.

This reduces the size of the sdist from 92,3KB to 33,1KB

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,8 +22,7 @@ package_dir =
 packages = find:
 install_requires =
     xopen >= 1.4.0
-setup_requires =
-    cython >= 0.29.20
+
 [options.packages.find]
 where = src
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,8 @@ package_dir =
 packages = find:
 install_requires =
     xopen >= 1.4.0
-
+setup_requires =
+    cython >= 0.29.20
 [options.packages.find]
 where = src
 
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,7 @@ from setuptools import setup, Extension
 import setuptools_scm  # noqa  Ensure it’s installed
 
 setup(
-    ext_modules=
-        [
-            Extension("dnaio._core", sources=["src/dnaio/_core.pyx"]),
-        ],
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -22,7 +22,8 @@ package_dir =
 packages = find:
 install_requires =
     xopen >= 1.4.0
-
+setup_requires =
+    cython >= 0.29.20
```
marcelm, line 26: This is redundant with Cython listed in `pyproject.toml`.

setup.py
```diff
@@ -1,11 +1,9 @@
 from setuptools import setup, Extension
-from Cython.Build import cythonize
 import setuptools_scm  # noqa  Ensure it’s installed
 
 setup(
-    ext_modules=cythonize(
+    ext_modules=
```
marcelm, line 5: Indentation is now messed up (you could run black on it once).

setup.cfg
rhpvorderman: Some versions of pip do not support the new package model yet, but in that case `setup_requires` is incomplete now. I will remove it.

marcelm: That would be *very* old versions of pip.

rhpvorderman:
All done.

marcelm:
Looks good!

PR #53, opened 2022-02-23
Rename (Bytes-)Sequence to (Bytes-)SequenceRecord

### LINKED ISSUE #36: Consider renaming Sequence to SequenceRecord
This has bothered me for a long time: A `Sequence` isn’t actually just a sequence; it is an object containing a sequence of nucleotides, a sequence of quality values and a name. The misnaming becomes obvious when one considers that a `Sequence` has a `sequence` attribute, so I’m sometimes tempted to write something like `s = sequence.sequence` or similar.

Biopython has `SeqRecord`, but I don’t like abbreviations in class names and to reduce the risk of being confusing with the Biopython class, I think it should be `SequenceRecord`.

marcelm:
Work in progress.

To Do:
- [x] Add the `Sequence` and `BytesSequence` aliases
- [x] Add tests for the aliases

For better testing, l’ll change this so that in the first step, the classes are renamed and aliases are added, but that the tests use the old names. Then I’ll change the tests to use the new names and add some tests for the old names.

Closes #36


marcelm:
I have added tests and moved all tests for the sequence classes to a separate module. Ready for review.

rhpvorderman (approved):
Looks good to me! One nitpick, `test_sequence.py` should probably be called `test_sequence_record.py`.

marcelm:
Since it also tests `BytesSequenceRecord`, I renamed it to `test_records.py` now.

rhpvorderman:
Great, much better naming!

PR #54, opened 2022-03-02
Bytes object touchups: no None, docstrings.

rhpvorderman:
Same as for Sequence. I added more stringent checks on BytesSequence getters and setters so None is not allowed for name and sequence.
I also copied and modified some docstrings.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -215,6 +215,20 @@ cdef class Sequence:
 
 
 cdef class BytesSequence:
+    """
+    A sequencing read with read name/id and (optional) qualities
+
+    If qualities are available, they are as
+    For a Sequence a FASTA file
+    record containing a read in a FASTA or FASTQ file. For FASTA, the qualities attribute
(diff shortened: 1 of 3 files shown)


src/dnaio/_core.pyx
```diff
@@ -215,28 +215,80 @@ cdef class Sequence:
...
 
 cdef class BytesSequence:
+    """
+    A sequencing read with read name/id and (optional) qualities
+
+    If qualities are available, they are as
+    For a Sequence a FASTA file
+    record containing a read in a FASTA or FASTQ file. For FASTA, the qualities attribute
+    is None. For FASTQ, qualities is a bytes object and it contains the qualities
+    encoded as ASCII(qual+33).
```
marcelm, line 225: That text is messed up. Just leave it out and I’ll fix it up in both Sequence and BytesSequence when I go through the docstrings.

rhpvorderman: Whoops. Shouldn't have copied that ad verbatim indeed. Removed that paragraph now.

PR #55, opened 2022-03-02
Remove unused imports. Make checks more consistent.

rhpvorderman:
+ Use PyBytes_CheckExact and PyUnicode_CheckExact everywhere (Cython also does this internally when declaring `str` or `bytes`).
+ Replace one instance of `strncmp` with `memcmp` (functionally equivalent in this particular case).
+ Use PyUnicode_IS_COMPACT_ASCII instead of PyUnicode_KIND checks in record names match. This is more strict, but consistent with the rest of the code base.

This is mostly some code tidying to remove some imports. Nothing critical.


marcelm:
Sorry, forgot about this one.

PR #58, opened 2022-03-08
Use Py_SIZE instead of PyBytes_GET_SIZE

rhpvorderman:
PyBytes_GET_SIZE performs a small check with 'assert' to ensure the object is a bytes object. Py_SIZE does not perform any checks and just gives the 'ob_size' attribute from the struct.

See: https://docs.python.org/3/c-api/structures.html?highlight=py_size#c.Py_SIZE

Also work for other PyVarObjects such as tuples. Widely used inside the CPython code base.

Not suitable for Unicode objects.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -1,9 +1,10 @@
 # cython: language_level=3, emit_code_comments=False
 
 from cpython.buffer cimport PyBUF_SIMPLE, PyObject_GetBuffer, PyBuffer_Release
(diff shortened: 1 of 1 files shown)


marcelm:
I’m fine with merging this if you think you want this for consistency or similar, but for me, extensions are compiled with `-DNDEBUG`, so the assertion check is already disabled anyway.

marcelm:
Or does Python use a different `assert` macro?

rhpvorderman:
No, it simply uses assert. 

> I’m fine with merging this if you think you want this for consistency or similar, but for me, extensions are compiled with -DNDEBUG, so the assertion check is already disabled anyway.

Ah, I did not know that. Then this PR makes no sense. Thanks. I will use that for my own extensions as well then!

PR #60, opened 2022-03-13
Move string_is_ascii to _update_buffer

### LINKED ISSUE #59: ASCII check should be moved to update buffer.
This way the check can be performed on large contiguous chunks of memory. Also it removes a branch from the __next__ method. Since the update buffer method is used relatively few times compared to the __next__ method it should be a net win.

Also because of the large contiguous chunks benefits of SIMD optimizations might become noticable. 

I cannot implement this right now, but I have to write this down otherwise it will be stuck in my head for the rest of the day.

rhpvorderman:
fixes #59 

Before:
```
  Time (mean ± σ):      1.001 s ±  0.018 s    [User: 884.8 ms, System: 115.9 ms]
  Range (min … max):    0.962 s …  1.045 s    50 runs
```
after:
```
  Time (mean ± σ):     930.0 ms ±  22.7 ms    [User: 819.8 ms, System: 110.1 ms]
  Range (min … max):   889.0 ms … 983.5 ms    50 runs
```

Quite interesting. So I commented out the checking code to get the performance without checking:
```
  Time (mean ± σ):     879.8 ms ±  24.0 ms    [User: 765.1 ms, System: 114.6 ms]
  Range (min … max):   842.2 ms … 944.9 ms    50 runs
```

So this is quite a significant reduction in overhead costs for the ASCII check.

The cost is that the record with non-ASCII character cannot be accurately reported anymore. But I expect that check to fail very rarely (if ever), so the speed improvement weighs heavier in my opinion.


rhpvorderman:
Good news. I tested  it against my SSE branch that does the simplest SIMD optimization and that removes the overhead entirely.

marcelm:
Cool, impressive work!

PR #28, opened 2022-01-13
Support Windows

marcelm:
- The GitHub Actions job that created the wheels needed to become a matrix job that runs on Windows and Linux. Because the sdist was also created in that job and I did not want it to be created multiple times, sdist creation, wheel creation and publishing are now separate jobs.
- `pipx` is installed on the Linux runners, so building the sdist could be simplified to `pipx run build` (and to test whether a wheel can be created from the sdist, I do not add the `--sdist` option).
- The other changes were just the things necessary to make the code work on both platforms.
- The Windows tests should no longer fail as soon as https://github.com/pycompression/xopen/pull/83 is merged and a new xopen version is published.

PR #63, opened 2022-03-18
Drop Python 3.6 support

marcelm:
Python 3.6 is end-of-life and there have recently been new releases of
Python 3.7+ that fix security issues:
https://discuss.python.org/t/14353

Python 3.6 did not get these fixes, so it seems a good idea to stop using
and/or supporting it.


rhpvorderman:
Python-isal will also not be able to use versions older than 3.7 due to `METH_FASTCALL` being introduced in 3.7. (When I made that PR 3.6 had been unsupported for a few weeks). So when I will release python-isal 1.0 everything still running 3.6 will break. I am all for this change!

Incidentally this also means the default settings of CIBuildWheel can be used as regards to the container used.

marcelm:
> Incidentally this also means the default settings of CIBuildWheel can be used as regards to the container used.

What do you mean?

rhpvorderman:
I was thinking of this https://github.com/marcelm/dnaio/pull/28#issuecomment-1013158275 . But I see now that refers to a version of pip, not to python 3.6. I misremembered, sorry.

marcelm:
Ah, but you’re right: Ubuntu 18.04 doesn’t have Python 3.6 either, so we give that up anyway and that means we can now bump the cibuildwheel version. I’ll add that to this PR.

rhpvorderman:
Okay great!

PR #68, opened 2022-03-30
Branchless checking for ASCII

rhpvorderman:
This is 20% faster for the ascii check. But 20% faster on something that is not a bottleneck is not really noticable. Still, I don't like wasting compute cycles and this makes the code simpler and without branches. 

I have been fiddling a bit with the ascii-check repo. I also added an SSE2 implementation in that repo. SSE2 in particular is a **very** interesting instruction set because **all** x86-64 CPUs support it. No exceptions. So it can always be included in the x86-64 builds. 

That implementation cuts the ASCII checking time in more than half, so that might be interesting to look at. I will add that later after this is merged.

### DIFF
--- a/src/dnaio/ascii_check.h
+++ b/src/dnaio/ascii_check.h
@@ -7,33 +7,28 @@
 static int
 string_is_ascii(char * string, size_t length) {
     size_t n = length;
+    uint64_t all_chars = 0;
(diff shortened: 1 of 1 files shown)


marcelm:
I forgot about this one.

PR #69, opened 2022-04-06
Avoid some "unclosed file" ResourceWarnings

marcelm:
This should fix the test failures in marcelm/cutadapt#609.

Possibly we should start to treat these warnings as errors just as in Cutadapt. I was a bit undisciplined with closing files in Cutadapt in the beginning, but then this actually led to truncated files that a user reported. Since then, I try to heed these warnings and close everything properly (so not relying on automatic cleanup at process termination).


rhpvorderman (approved):

rhpvorderman:
Agreed. And in case dnaio is used in some server process (unlikely but still) the application may never be closed at all.

PR #71, opened 2022-04-08
Use Black for formatting

marcelm:
This should be self-explaining.

I also moved the helper script to `helpers/` so that I could avoid listing the file name explicitly in `tox.ini`.


rhpvorderman:
> This should be self-explaining.

But let's have a 300+ comment discussion on the merits of `)` before or after a newline regardless :wink:.

marcelm:
OMG, in the first second after reading your comment, I didn’t notice the smiley ...

Thanks for cutting the discussion short :smiley:

PR #72, opened 2022-04-08
Add a little tutorial

marcelm:


doc/tutorial.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    with dnaio.open("reads.fastq.gz") as reader:
+        bp = 0
+        for record in reader:
+            bp += len(record)
+    print(f"The input file contains {bp/1E6:.1f} Mbp")
+
+As can be seen from the ``.gz`` file extension,
+the input file is gzip-compressed.
+`dnaio.open` detects and handles this automatically by opening the file with
```
rhpvorderman, line 26: Using single backticks creates referencable text, but not code. Double backticks are needed in the latter case.

Do these references reference back to the API documentation?

doc/tutorial.rst
```diff
@@ -0,0 +1,131 @@
...
+`xopen <https://github.com/pycompression/xopen/>`_.
+
+Here, the call to `dnaio.open` returns a `FastqReader` object.
+Iterating over it in the ``for`` loop results in `SequenceRecord` objects.
```
rhpvorderman, line 30: On the next line double backticks are used for SequenceRecord. Was this intentional?

rhpvorderman (commented):
Looks good to me, excellent tutorial. I posted a few minor gripes.

doc/tutorial.rst
```diff
@@ -0,0 +1,131 @@
...
+The only essential concepts to know about are
+the `dnaio.open` function and the `SequenceRecord` object.
+
+
+Reading
+-------
+
+The main interface for reading and writing sequence files is the `dnaio.open` function.
+For example, this program reads in a FASTQ file and computes the total number of nucleotides
+it contains::
```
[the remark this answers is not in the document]
marcelm: I haven’t seen this file rendered, but the example from the README (which does not use `code-block`) is syntax highlighted: https://dnaio.readthedocs.io/en/latest/#example-usage

doc/tutorial.rst
marcelm: Yes, the single backticks were intentional. It worked for `dnaio.open` in the README, but now I see that it doesn’t work for `SequenceRecord` and all the other classes where I don’t have a `dnaio.` prefix.

doc/tutorial.rst
marcelm: I wanted to create a link only for the first occurrence to avoid too much blue in the text, so this is intentional. (Apart from the fact that it doesn’t work as I discovered now.)

doc/tutorial.rst
marcelm: Ah, I need to write `~dnaio.SequenceRecord` to get what I want.

marcelm:
I fixed the nonworking links.

rhpvorderman:
Excellent!

PR #74, opened 2022-04-15
Vectorize the ASCII check using SSE2 instructions

rhpvorderman:
SSE2 is guaranteed to be present on al AMD64 (x86_64) platforms. So a simple check for such a platform is sufficient to enable the instruction set without running into compile problems.

This increases the ASCII check speed from 20GB/s to 50GB/s. Making our ASCII string cost creation almost free.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,21 @@
+import platform
+import sys
+
 from setuptools import setup, Extension
 import setuptools_scm  # noqa  Ensure it’s installed
 
+# Check for 64-bit
+if sys.maxsize > 2**32 and (
+    platform.machine() == "x86_64" or platform.machine() == "AMD64"
+):
+    DEFINE_MACROS = [("USE_SSE2", None)]
+else:
+    DEFINE_MACROS = []
+
 setup(
     ext_modules=[
-        Extension("dnaio._core", sources=["src/dnaio/_core.pyx"]),
+        Extension(
+            "dnaio._core", sources=["src/dnaio/_core.pyx"], define_macros=DEFINE_MACROS
+        ),
     ],
 )
--- a/setup.py
+++ b/setup.py
@@ -5,9 +5,7 @@ from setuptools import setup, Extension
 import setuptools_scm  # noqa  Ensure it’s installed
 
 # Check for 64-bit
-if sys.maxsize > 2**32 and (
-    platform.machine() == "x86_64" or platform.machine() == "AMD64"
-):
+if platform.machine() == "x86_64" or platform.machine() == "AMD64":
     DEFINE_MACROS = [("USE_SSE2", None)]
 else:
     DEFINE_MACROS = []
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,6 @@ import sys
 from setuptools import setup, Extension
 import setuptools_scm  # noqa  Ensure it’s installed
 
-# Check for 64-bit
 if platform.machine() == "x86_64" or platform.machine() == "AMD64":
     DEFINE_MACROS = [("USE_SSE2", None)]
 else:
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -14,7 +14,14 @@ cdef extern from "Python.h":
     bint PyUnicode_IS_COMPACT_ASCII(object o)
     object PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar)
 
-cdef extern from "ascii_check.h":
+cdef extern from *:
+    """
+    #if defined(USE_SSE2)
+      #include "ascii_check_sse2.h"
+    #else
+      #include "ascii_check.h"
+    #endif
+    """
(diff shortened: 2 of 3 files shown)


src/dnaio/_core.pyx
```diff
@@ -14,7 +14,14 @@ cdef extern from "Python.h":
     bint PyUnicode_IS_COMPACT_ASCII(object o)
     object PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar)
 
-cdef extern from "ascii_check.h":
+cdef extern from *:
+    """
+    #if defined(USE_SSE2)
```
marcelm, line 19: You could use `#if defined(__x86_64__) || _M_X64` directly and then leave `setup.py` unchanged. `__x86_64__` is for GCC and Clang and `_M_X64` is for Visual Studio, see http://jdebp.info/FGA/predefined-macros-processor.html .

rhpvorderman: The advantage of this way is that it makes it also quite easy to test if the non-sse2 version is correct by undefining the USE_SSE2 macro.
I am aware that compiler-specific options exist, but to me that does not look like a particularly stable, reliable and universal API to build on.

setup.py
```diff
@@ -1,8 +1,21 @@
+import platform
+import sys
+
 from setuptools import setup, Extension
 import setuptools_scm  # noqa  Ensure it’s installed
 
+# Check for 64-bit
+if sys.maxsize > 2**32 and (
```
marcelm, line 8: If `platform.machine()` already says "x86_64", it should not be necessary to check whether the platform is 64 bit.

src/dnaio/_core.pyx
marcelm: I agree with your first paragraph. Testing is a valid reason, absolutely, so I’m fine with using that option.

Regarding your other point, I disagree. Querying a documented, pre-defined compiler macro is totally fine and in my opinion not worse than relying on `platform.machine()`, which is at the same level of "universalness", demonstrated by having to check for both `"x86_64"` and `"AMD64"`.

marcelm: isa-l uses this in at least one of their files:
https://github.com/intel/isa-l/blob/ebb78fc99ef7247bd9c16cecf7faa248563ce4f6/erasure_code/ec_highlevel_func.c#L32

rhpvorderman:
> Regarding your other point, I disagree. Querying a documented, pre-defined compiler macro is totally fine and in my opinion not worse than relying on platform.machine(), which is at the same level of "universalness", demonstrated by having to check for both "x86_64" and "AMD64".

At least with `platform.machine` there are only two options. x86_64 and AMD64. There are quite a lot of C compilers out there. GCC, Clang, MSVC, Intel C compiler, AMD Optimizing Compiler etc. So there is definitely going to be more variety in pre-defined compiler macros. There is no standardization in this space at all, so I feel the `platform.machine` choice is safer. 

Point taken for the "documented" part. The macros should be at at least as stable as the platform.machine option.

setup.py
rhpvorderman: Agreed. Done!

src/dnaio/ascii_check_sse2.h
```diff
@@ -0,0 +1,67 @@
...
+        n -= sizeof(longword);
+    }
+    char_ptr = (char *)longword_ptr;
+    while (n != 0) {
+        all_chars |= *char_ptr;
+        char_ptr += 1;
+        n -= 1;
+    }
+    // Check the most significant bits in the accumulated words and chars.
+    return !(_mm_movemask_epi8(all_words) || (all_chars & ASCII_MASK_1BYTE));
```
marcelm, line 66: Nice how the movemask instruction is such a good fit here.

marcelm:
Looks good now – although I am a bit disappointed that the `setup.py` isn’t so nice and short anymore ...

Thanks!

src/dnaio/ascii_check_sse2.h
rhpvorderman: It is a very useful instruction. Intrinsic compare functions set the most significant bit too. So if you compare one vector to another you end up with a vector of bytes with the most significant bit set. There is also a popcnt (POPCOUNT) instruction that simply reports the number of set bits. So you can use mm_cmpneq + mm_movemask + popcnt to calculate the hamming distance of a vector in just three instructions.

There is also mm_blend, where  you create a new vector from two other vectors, based on a provided mask. Very useful, as this allows branchless programming while still using conditionals (create a mask with a compare function, calculate the two possible result vectors, then select based on the mask). They use this in minimap2 for the alignment algorithm. So that might be interesting for cutadapt.

rhpvorderman:
> Looks good now – although I am a bit disappointed that the setup.py isn’t so nice and short anymore ...

The sacrifices we make for a few % performance gains... What have we become?!

If it makes you feel better you can take a look at the python-isal setup.py ;-). Although that has become slightly less verbose with the move to a pure C extension.

PR #75, opened 2022-04-19
Small pointer refactor and cleanup

rhpvorderman:
This removes some redundant variables as well as making quite a few lines of code shorter. It also reduces the amount of pointer arithmetic. (Much less `self.buffer + some_offset`.) 

Additionally all the "pointer slicing" (was exclusively used in errors) has been replaced with the appropriate calls to PyUnicode_DecodeASCII.

PyUnicode_1BYTE_DATA has been replaced with PyUnicode_DATA. Since we cast the results to `<char *>` anyway
 there is no need for the macro that casts to `Py_UCS1`. 

I also inlined several functions that were previously shared between SequenceRecord and BytesSequenceRecord. Which is now redundant.

I think this makes the code more readable.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -392,7 +392,7 @@ cdef class FastqIter:
         bint yielded_two_headers
         bint eof
         object file
-        Py_ssize_t record_start
+        char *record_start
     cdef readonly Py_ssize_t number_of_records
 
     def __cinit__(self, file, sequence_class, Py_ssize_t buffer_size):
@@ -407,7 +407,7 @@ cdef class FastqIter:
         self.extra_newline = False
         self.yielded_two_headers = False
         self.eof = False
-        self.record_start = 0
+        self.record_start = self.buffer
         self.file = file
         if buffer_size < 1:
             raise ValueError("Starting buffer size too small")
@@ -424,7 +424,7 @@ cdef class FastqIter:
         cdef char *tmp
         cdef Py_ssize_t remaining_bytes
 
-        if self.record_start == 0 and self.bytes_in_buffer == self.buffer_size:
+        if self.record_start == self.buffer and self.bytes_in_buffer == self.buffer_size:
             # buffer too small, double it
             self.buffer_size *= 2
             tmp = <char *>PyMem_Realloc(self.buffer, self.buffer_size)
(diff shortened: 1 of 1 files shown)


src/dnaio/_core.pyx
```diff
@@ -433,11 +401,11 @@ cdef class FastqIter:
...
         else:
             # Move the incomplete record from the end of the buffer to the beginning.
-            remaining_bytes = self.bytes_in_buffer - self.record_start
+            remaining_bytes = self.bytes_in_buffer - (self.record_start - self.buffer)
             # Memmove copies safely when dest and src overlap.
-            memmove(self.buffer, self.buffer + self.record_start, remaining_bytes)
+            memmove(self.buffer, self.record_start, remaining_bytes)
             self.bytes_in_buffer = remaining_bytes
-            self.record_start = 0
+        self.record_start = self.buffer
```
marcelm, line 408: Did you de-dent this on purpose?

src/dnaio/_core.pyx
```diff
@@ -485,58 +452,57 @@ cdef class FastqIter:
...
             object ret_val
-            Py_ssize_t name_start, name_end, name_length
-            Py_ssize_t sequence_start, sequence_end, sequence_length
-            Py_ssize_t second_header_start, second_header_end, second_header_length
-            Py_ssize_t qualities_start, qualities_end, qualities_length
-            char *name_end_ptr
-            char *sequence_end_ptr
-            char *second_header_end_ptr
-            char *qualities_end_ptr
+            char * name_start
```
marcelm, line 455: I know you have consistently used this style in your PRs whenever you changed or added any pointer declarations, but I think this is not an improvement. There [are arguments for both the `char* p` and `char *p` style](https://www.stroustrup.com/bs_faq2.html#whitespace), but using spaces on both sides of the `*` is the option that seems to be the most unconventional. CPython itself seems to use `char *p` most of the time.

src/dnaio/_core.pyx
```diff
@@ -547,26 +513,25 @@ cdef class FastqIter:
             qualities_length = qualities_end - qualities_start
 
             # Check for \r\n line-endings and compensate
-            if self.buffer[name_end - 1] == b'\r':
+            if (name_end - 1)[0] == b'\r':
```
marcelm, line 516: I think switching to only pointers instead of indices is an improvement overall here, but I want to mention one downside as well: Previously, it was clear that `self.buffer[name_end - 1]` accesses an element of the buffer, and now that’s not the case anymore. One has to know from the context that `name_end` points into the buffer.

Being able to drop the `name_end_ptr`/`name_end` duality offsets this, though, IMO.

marcelm (commented):
A couple of comments, as always.

src/dnaio/_core.pyx
rhpvorderman: Yes. There are two blocks. The above block resizes buffer. This means that `self.buffer` can point to a different memory adress.

Since self.record_start is now a pointer itself rather than an offset, it needs to be set to the start of the buffer (self.buffer) in both cases.

src/dnaio/_core.pyx
rhpvorderman: In my mind I read it as "char pointer name_start" hence the spaces. But I will conform to the `char *p` style. Sorry I forgot about this.

src/dnaio/_core.pyx
rhpvorderman: Yes, I also think this is uglier than what was used previously. Luckily it still reads quite okay: one address back from the end of the name string.

src/dnaio/_core.pyx
rhpvorderman: Done

marcelm:
Thanks!

PR #80, opened 2022-05-17
Eliminate extra type checking loop in records_are_mates

rhpvorderman:
This does not improve the speed (which was of course the original goal). However, the code looks nicer so I decided to push. It eliminates a loop and it eliminates a branch. Review at your leisure ;-).

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -664,14 +664,11 @@ def records_are_mates(*args):
     if args_length < 2:
         raise TypeError("records_are_mates requires at least two arguments")
 
-    cdef Py_ssize_t i
-    cdef object item
-    for i in range(args_length):
-        item = <object>PyTuple_GET_ITEM(args, i)
-        if Py_TYPE(item) != <PyTypeObject *>SequenceRecord:
-            raise TypeError(f"{item:r} is not a SequenceRecord object")
+    cdef SequenceRecord first = <SequenceRecord>PyTuple_GET_ITEM(args, 0)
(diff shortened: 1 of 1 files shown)


src/dnaio/_core.pyx
```diff
@@ -683,17 +680,17 @@ def records_are_mates(*args):
...
         other_name_length = PyUnicode_GET_LENGTH(other_name_obj)
-        are_mates = record_ids_match(first_name, other_name, id_length,
-                                     other_name_length, id_ends_with_number)
-        if not are_mates:
-            return False
-
-    # All sequence records checked, no mismatches found.
-    return True
+        # If a match is false, are_mates will stay false regardless of any true checks afterward.
+        are_mates &= record_ids_match(first_name, other_name, id_length,
```
marcelm, line 694: This is less efficient because it now always processes all records (no short-circuit evaluation anymore). Before, it would stop as soon as there was a mismatch. The expected case is that the record IDs are all identical, so it shouldn’t matter, but still ...

rhpvorderman: This is less efficient when we take the case of a single group of records with a mismatch in isolation. However, the most common case is that records are mates. And it is faster in that case because the branch is eliminated. This means we will reach the cause of error sooner rather than later in most cases. Since we process all correct trios that come before it faster.

Also, one loop takes just 8 nanoseconds. That is faster than your screen refresh delay by a factor of ... *2 MILLION* so I wouldn't worry too much about the error being thrown slightly later for the user to debug. (I agree that is an important benchmarking metric!)

PR #81, opened 2022-05-27
API documentation improvements (and some other minor things)

### LINKED ISSUE #56: API review
Here are a couple of things I noticed should possibly be revised before releasing a version 1.0. I’ll extend this when I find more.

* `SequenceRecord.fastq_bytes_two_headers()` is redundant with `SequenceRecord.fastq_bytes(two_headers=True)`
* `record_names_match` and `record_names_match_bytes` should possibly be renamed to `record_ids_match` and `record_ids_match_bytes` because those names are more accurate (also requires that the C-level `record_ids_match` is renamed)

marcelm:
A couple of cleanup commits:
- Document Reader and Writer interfaces and `records_are_mates`
- Some API documentation cleanup
- Factor out `_open_paired`
- Type hint corrections

I think this should close #56.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -63,7 +63,7 @@ which has a `~dnaio.FastqWriter.write()` method that accepts a ``SequenceRecord`
 
 Instead of constructing a single record from scratch,
 in practice it is more realistic to take input reads,
-process them somehow and write them to a new output file.
+process them, and write them to a new output file.
 The following example program shows how that can be done.
 It truncates all reads in the input file to a length of 30 nt
 and writes them to another file::
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -3,7 +3,7 @@ Tutorial
 
 This should get you started with using ``dnaio``.
 The only essential concepts to know about are
-the `dnaio.open` function and the `SequenceRecord` object.
+the `dnaio.open` function and the `~dnaio.SequenceRecord` object.
 
 
 Reading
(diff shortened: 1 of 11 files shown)


doc/tutorial.rst
```diff
@@ -59,12 +59,13 @@ pass the ``mode="w"`` argument to ``dnaio.open``::
         writer.write(dnaio.SequenceRecord("name", "ACGT", "#B!#"))
 
 Here, a `~dnaio.FastqWriter` object is returned by ``dnaio.open``,
-which has a ``~dnaio.FastqWriter.write()`` method that accepts a ``SequenceRecord``.
+which has a `~dnaio.FastqWriter.write()` method that accepts a ``SequenceRecord``.
 
 Instead of constructing a single record from scratch,
-it may be more realistic to take input reads,
+in practice it is more realistic to take input reads,
 process them somehow and write them to a new output file.
```
rhpvorderman, line 66: I would also get rid of "somehow" here as it does not add any meaning.

src/dnaio/_core.pyx
```diff
@@ -239,10 +239,7 @@ cdef class SequenceRecord:
 
 
     def fastq_bytes_two_headers(self):
-        """
-        Return this record in FASTQ format as a bytes object where the header (after the @) is
-        repeated on the third line.
-        """
+        # Deprecated, use ``.fastq_bytes(two_headers=True)`` instead.
```
rhpvorderman, line 242: should we issue a DeprecationWarning here? (Similar how to CPython tackles these).

rhpvorderman (approved):
Looks good to me! I added two minor non-blocking comments.

src/dnaio/_core.pyx
rhpvorderman: Sure. I doubt people are going to complain. Who wants bigger FASTQ data anyway?

PR #86, opened 2022-07-15
Version 0.9.1

rhpvorderman:
Hi, I tried to release dnaio on my own since you are on holiday. It seems as I don't have permissions though. (I wanted to push to main and add a v0.9.1 tag). No matter. I think I can help the user by providing the wheel directly. I made this PR so the changelog work does not need to be duplicated. See you after the holiday!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,10 +2,10 @@
 Changelog
 =========
 
-v0.9.1 (2022-07-15)
+v0.9.1 (2022-08-01)
 -------------------
 
-* :pr:`85`: Mac OS wheels are now also build as part of the release procedure.
+* :pr:`85`: macOS wheels are now also built as part of the release procedure.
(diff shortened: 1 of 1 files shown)


marcelm:
I don’t think I can grant any more permissions (for repositories under personal accounts, permissions are not very fine-grained. It’s just 'Add collaborator'). What error did you get? I’ll make the release now.

marcelm:
Not that it should be relevant for this particular problem, but I invited you to the PyPI project as well (just in case).

rhpvorderman:
I could not push the tag to remote (no permissions). Which was quite essential. And I cannot access the settings to change it. (I have no settings access at all). 
I figured that in this particular case you wouldn't mind if I used my administrative rights to solve a small issue for a user with minimal or no changes to the codebase, but it turned out I didn't have any ;-).

Issue #82, opened 2022-06-01
Switch single&paired API to single&multiple API.


rhpvorderman:
Currently I am working a lot with UMI data that is stored in a separate FASTQ file meaning I have 3 files now.

I needed to filter those files on average error rate so I adopted the [fastq-filter](https://github.com/LUMC/fastq-filter) program to work with multiple files.

To keep the pipeline simple. I opted to have a Multiple file reader. This yields 1-tuples for 1 file, 2-tuples for 2 files, 3-tuples for 3 files, etc. 
This way I can write the filters to always handle a tuple of SequenceRecord objects and use the same filter in all cases.
Similarly I wrote a multiple writer.

I am wondering if we should do this in dnaio too. There are now two cases in dniao:
1. Single file. Yield one SequenceRecord object.
2. Paired file. Yield a 2-tuple of SequenceRecord objects.

I propose replacing the latter with a multipe file reader that can read n number of records and yields n-tuples of SequenceRecords. The PairedEndReader and PairedEndWriter interfaces can still be maintained, but these can simply inherit the MultipleReaders and provide a backwards compatible interface. (Shouldn't be too hard given it is just the `2`-case of the MultipleReader). 

This way I do not have to reinvent the wheel across multiple projects. I also feel this is needed for cutadapt. Which needs a sort of auxilary file option, where the auxilary file with the UMIs is kept in sync with the FASTQ files that are output from cutadapt. Currently I have to use biopet-fastqsync to sync the UMI FASTQ file afterwards. (This is not the correct place to raise this issue, but I simply state this here to show that I think this will be a good move for the future).

I already have implemented a multiple reader in my FASTQ filter project. At first it was written in a generic manner. (Everything is a list of multiple files.) But I discovered that severely harms the single-end and paired-end cases: https://github.com/LUMC/fastq-filter/pull/16 . I wonder what the best way is to implement is in dnaio. Alternatively there could be separate 1-tuple 2-tuple n-tuple readers that all share the same interface trough abstract classes.

marcelm:
Generalizing the paired-end reader to multiple files sounds like a good idea. I think I’d implement this by accepting more than two input files in `dnaio.open` and then the function would work as before for n=1 and n=2 (so totally backwards compatible for the single end and paired-end cases). Then for n>2, it would return this new MultipleReader (not sure whether that is the best name, though). Is that what you meant?

This would indeed be a requirement for supporting records with more than two "ends" in Cutadapt.

[...]

PR #87, opened 2022-07-19
Multiple reader/writer api

rhpvorderman:
I added the ability for `dnaio.open` to open multiple files.  I also added the `dnaio.open_multiple` API which iworks in case you want 1-tuples for 1 file. (I use this sort of code in my fastq filter and deduplication programs)

dnaio.open is backwards compatible. Also I think `MultipleFastqWriter` is probably going to be faster than `PairedFastqWriter`, but I will benchmark that later and backport any changes.

The `MultipleFastXWriter` objects also have a `write_iterable` call which you can give a SequenceRecord tuple generator. This is much faster than repeatedly calling write calls in a for loop. This can also be backported to the PairedFastqWriter and SingleFastqWriter at a later date.

The API is fully tested. Just need to add some docstrings. Posting this here in advance for code review and to make sure this effort is not duplicated.


rhpvorderman:
on second thought, this extra open_multiple API is going to be confusing and there is some code duplication. Let me simplify this.

marcelm:
Nice, I’ve started a review and have gotten halfway, will submit when I’m finished.

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,205 @@
...
+        for writer in self.writers:
+            writer.close()
+
+    def write(self, *records: SequenceRecord):
+        if len(records) != self.number_of_files:
+            raise ValueError(f"records must have length {self.number_of_files}")
+        for record, writer in zip(records, self.writers):
+            writer.write(record.fastq_bytes())
+
+    def write_iterable(self, records_iterable: Iterable[Tuple[SequenceRecord, ...]]):
```
marcelm, line 173: This needs a docstring that describes when to use this method over `write()`.

I’m not so fond of the name; would something like `write_many` be clearer? (Not perfect either, though.)

tests/test_multiple.py
```diff
@@ -0,0 +1,121 @@
...
+    records = [SequenceRecord("A", "A", "A") for _ in range(number_of_files)]
+    with _open_multiple(*files, mode="w", fileformat=fileformat) as writer:
+        writer.write(*records)
+
+
+@pytest.mark.parametrize(
+    ["number_of_files", "fileformat"],
+    itertools.product((1, 2, 3, 4), ("fastq", "fasta")),
+)
+def test_multiple_write_too_much(number_of_files, fileformat):
```
marcelm, line 73: Should be `_too_many`

marcelm (commented):
I think this is a good addition. I am a little bit concerned, though, that this is code that will see much less use (and therefore testing) than the code for the 1- and 2-file cases because there’s presumably very little data out there that needs this. What would be the downside of not only backporting some of this to the PairedEnd... classes, but merging them (at some point in the future, not now) so that there’s only one way to read more than one file?

A note regarding terminology: I noticed in the code that the variable name "records" was used quite often, but it’s now less clear whether that refers to records belonging to the same molecule fragment (forward and reverse read in the case of paired-end reads) or whether it refers to multiple unrelated single ends. That is, if you imagine the data as a table, "records" could refer to a row or a column. I’m not sure what to do about this. Perhaps it’s clear enough from the context.

marcelm:
> What would be the downside of not only backporting some of this to the PairedEnd... classes, but merging them (at some point in the future, not now) so that there’s only one way to read more than one file?

It appears I forgot the discussion we had in #82.

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,205 @@
...
+    fileformat: Optional[str] = None,
+    mode: str = "r",
+    qualities: Optional[bool] = None,
+    opener=xopen,
+):
+    if mode not in ("r", "w", "a"):
+        raise ValueError("mode must be one of 'r', 'w', 'a'")
+    if mode == "r":
+        return MultipleFileReader(*files, fileformat=fileformat, opener=opener)
+    append = True if mode == "a" else False
```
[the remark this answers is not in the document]
rhpvorderman: :man_facepalming:

src/dnaio/multipleend.py
rhpvorderman: write_multiple? 
write_list?
Write list sounds quite okay, and properly conveys that the list can also have one entry. That it accepts any iterable container type should not really be that important.

marcelm: Hm, I think `write_iterable` is still better than those alternatives. Perhaps the method just needs a docstring and then it’s fine.

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,205 @@
...
+    if fileformat == "fastq" or qualities or (fileformat is None and qualities is None):
```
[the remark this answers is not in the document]
rhpvorderman: I added format detection. I assume the first file has the same type as the others. I will bet a beer (or other beverage) on it that this assumption will not lead to an issue on our bug tracker in the coming 10 years.
No need to add extra code to catch every corner case.

src/dnaio/__init__.py
```diff
@@ -122,11 +122,20 @@ def open(
...
        `PairedEndWriter`.
     """
-    if mode not in ("r", "rb", "w", "a"):
-        raise ValueError("Mode must be 'r', 'rb', 'w' or 'a'")
-    elif interleaved and file2 is not None:
-        raise ValueError("When interleaved is True, file2 must be None")
-    elif interleaved or file2 is not None:
+    if mode not in ("r", "w", "a"):
+        raise ValueError("Mode must be 'r', 'w' or 'a'")
+    if files and file2 is not None:
```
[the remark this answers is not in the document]
rhpvorderman: Agreed. Fixed it.

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,205 @@
...
+        self,
+        *files: Union[str, PathLike, BinaryIO],
+        fileformat: Optional[str] = None,
+        opener=xopen,
+    ):
+        if len(files) < 1:
+            raise ValueError("At least one file is required")
+        self.files = files
+        self.readers: List[Union[FastaReader, FastqReader]] = [
+            _open_single(  # type: ignore
```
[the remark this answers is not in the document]
rhpvorderman: Good catch. Implemented

src/dnaio/multipleend.py
rhpvorderman: The method has a docstring trough the MultipleFileWriter interface. It is an abstractmethod that is inherited.

src/dnaio/__init__.py
```diff
@@ -69,7 +70,7 @@
 
 def open(
     file1: Union[str, PathLike, BinaryIO],
-    *,
+    *files: Union[str, PathLike, BinaryIO],
     file2: Optional[Union[str, PathLike, BinaryIO]] = None,
```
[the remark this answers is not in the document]
rhpvorderman: technically `file1` should also be deprecated. This will make the interface much easier ->
```python3
dnaio.open("my.fastq", **kwargs) -> SingleEnd
dnaio.open("R1.fastq", "R2.fastq", **kwargs) -> PairedEnd
dnaio.open("R1.fastq", "R2.fastq", "R3.fastq", **kwargs) -> MultipleEnd
```
It will also become easier for us to use typing.overload and provide the correct typing information.

But in order to be backwards-compatible, file1 and file2 are kept intact.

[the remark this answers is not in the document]
rhpvorderman: Hmm. That is a good idea actually. I will remake the interface so that `*files` is the primary way of entring files. File1 and file2 should be deprecated keyword arguments. I will change the logic to accomodate this. This will also make it much easier to write the docstring.

rhpvorderman:
Done! Sorry for letting this one rest a bit longer. We are changing a clinical pipeline in our institute and that takes up most of my time.
Also ironically our sequence center has decided to use new machines/software where instead of providing an extra UMI sequence file, they can put it in the header. So this PR suddenly got less relevant. On the other hand, we do have `3` hardcoded multiple times in our code because it is a thing.

Documenting the *files parameter was hard until I decided to also deprecate `file1`. Now it is very straight forward. I just needed to add an elif tree to ensure backwards compatibility but at least the interface should be clear going forward. (If you agree, that is.)

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
+        )
+    elif mode not in ("r", "w", "a"):
+        raise ValueError("Mode must be 'r', 'w' or 'a'")
+
+    if interleaved or len(files) == 2:
         return _open_paired(
-            file1,
-            file2=file2,
+            files[0],
+            file2=None if interleaved else files[1],
```
rhpvorderman, line 162: I see some redundancy here in open paired. Either you give one or two files. One means interleaved, two means not. The interleaved parameter is this not necessary for the internal _open_paired function. It will be much easier simply to use `_open_paired(*files, **other_kwargs` since we have  already established above that the correct conditions are met.

marcelm: I agree, this could be simpler.

src/dnaio/__init__.py
```diff
@@ -68,31 +69,39 @@
...
+    MultipleFileReader,
+    MultipleFileWriter,
+]:
     """
     Open one or two files in FASTA or FASTQ format for reading or writing.
 
     Parameters:
+      files:
+        one or multiple Path or an open file-like objects. One for single-end
+        reads, two for multiple-end reads etc. More than two files are also
```
marcelm, line 94:
```suggestion
        one or more Path or open file-like objects. One for single-end
        reads, two for paired-end reads etc. More than two files are also
```

src/dnaio/__init__.py
```diff
@@ -68,31 +69,39 @@
...
 
       file2:
-        Path or an open file-like object. When reading paired-end reads from
-        two files, set this to the second file.
+        Deprecated keyword argument for the second file.
 
       mode:
-        Either ``'r'`` or ``'rb'`` for reading, ``'w'`` for writing
-        or ``'a'`` for appending.
+        Either ``'r'``, ``'w'`` for writing or ``'a'`` for appending.
```
marcelm, line 104:
```suggestion
        Set to ``'r'`` for reading, ``'w'`` for writing or ``'a'`` for appending.
```

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
        `PairedEndWriter`.
     """
-    if mode not in ("r", "rb", "w", "a"):
-        raise ValueError("Mode must be 'r', 'rb', 'w' or 'a'")
-    elif interleaved and file2 is not None:
-        raise ValueError("When interleaved is True, file2 must be None")
-    elif interleaved or file2 is not None:
+    if files and (file1 is not None):
+        raise ValueError(
+            "the file1 argument cannot be used together with unnamed files"
```
marcelm, line 136:
```suggestion
            "The file1 keyword argument cannot be used together with files specified"
            "as positional arguments"
```
(I initially associated "unnamed file" with something filesystem related.)

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
+        )
+    elif len(files) > 1 and file2 is not None:
+        raise ValueError(
+            "the file2 argument cannot be used together with more than one "
+            "unnamed file"
```
marcelm, line 141:
```suggestion
            "file specified as positional argument"
```

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
+        )
+    elif file1 is not None and file2 is not None and files:
+        raise ValueError(
+            "file1 and file2 arguments cannot be used together with unnamed files"
```
marcelm, line 145:
```suggestion
            "file1 and file2 arguments cannot be used together with files specified"
            "as positional arguments"
```

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
+        )
+    elif file1 is not None and file2 is not None:
+        files = (file1, file2)
+    elif file2 is not None and len(files) == 1:
+        files = (files[0], file2)
+
+    if len(files) > 1 and interleaved:
+        raise ValueError(
+            "When interleaved is True only one file must be specified as input."
```
marcelm, line 154:
```suggestion
            "When interleaved is True, only one file must be specified."
```

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
-            file2=file2,
+            files[0],
+            file2=None if interleaved else files[1],
             opener=opener,
             fileformat=fileformat,
             interleaved=interleaved,
             mode=mode,
             qualities=qualities,
         )
+    elif len(files) > 2:  # 3 or more files
```
marcelm, line 169:
```suggestion
    elif len(files) > 2:
```

src/dnaio/interfaces.py
```diff
@@ -40,3 +43,28 @@ def write(self, record1: SequenceRecord, record2: SequenceRecord) -> None:
...
+        Write N SequenceRecords to the output. N must be equal
+        to the number of files the MultipleFileWriter was initialized with.
+
+        This method does not check whether the records are properly paired.
+        """
+
+    @abstractmethod
+    def write_iterable(self, list_of_records: Iterable[Tuple[SequenceRecord, ...]]):
+        """
+        Iterates over the list (or other iterable container) and writes all
```
marcelm, line 63:
```suggestion
        Iterate over the list (or other iterable container) and write all
```

src/dnaio/__init__.py
```diff
@@ -122,21 +131,51 @@ def open(
...
-    elif interleaved and file2 is not None:
-        raise ValueError("When interleaved is True, file2 must be None")
-    elif interleaved or file2 is not None:
+    if files and (file1 is not None):
+        raise ValueError(
+            "the file1 argument cannot be used together with unnamed files"
+        )
+    elif len(files) > 1 and file2 is not None:
+        raise ValueError(
+            "the file2 argument cannot be used together with more than one "
```
marcelm, line 140:
```suggestion
            "The file2 argument cannot be used together with more than one "
```

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _open_multiple(
+    *files: Union[str, PathLike, BinaryIO],
+    fileformat: Optional[str] = None,
+    mode: str = "r",
+    qualities: Optional[bool] = None,
+    opener=xopen,
+):
+    if not files:
+        raise ValueError("at least one file is required")
```
marcelm, line 23:
```suggestion
        raise ValueError("At least one file is required")
```

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,221 @@
...
+    if mode not in ("r", "w", "a"):
+        raise ValueError("mode must be one of 'r', 'w', 'a'")
```
marcelm, line 25:
```suggestion
        raise ValueError("Mode must be one of 'r', 'w', 'a'")
```

src/dnaio/multipleend.py
```diff
@@ -0,0 +1,221 @@
...
+    elif mode == "r":
+        return MultipleFileReader(*files, fileformat=fileformat, opener=opener)
+    elif mode == "w" and fileformat is None:
+        # Assume mixed files will not be offered.
+        for file in files:
+            if isinstance(file, str):
```
marcelm, line 31: Should this also be made to work if `file` is `PathLike`?

tests/test_open.py
```diff
@@ -301,3 +302,50 @@ def test_islice_gzip_does_not_fail(tmp_path):
...
+def test_unsupported_mode():
+    with pytest.raises(ValueError) as error:
+        _ = dnaio.open(os.devnull, mode="x")
+    error.match("Mode must be")
+
+
+def test_no_file2_with_multiple_args():
+    with pytest.raises(ValueError) as error:
+        _ = dnaio.open(os.devnull, os.devnull, file2=os.devnull)
+    error.match("more than one unnamed file")
```
marcelm, line 316: Also needs to be adjusted if you accept my suggestion about changing the error message

marcelm (commented):
Sorry about the very long delay. Are you still interested in merging this? I think your most recent change of deprecating the `file1` and `file2` keyword arguments and instead just having a `*files` positional argument is a very nice improvement.

I have added a couple of suggestions, but they are mostly nitpicks about spelling.

src/dnaio/__init__.py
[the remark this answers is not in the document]
marcelm: I like that!

src/dnaio/multipleend.py
rhpvorderman: That is quite an easy thing. Done.

rhpvorderman:
> Sorry about the very long delay. Are you still interested in merging this? 

Yes. Aren't you? :wink: https://github.com/marcelm/cutadapt/issues/491#issuecomment-1274754051 
I have two applications (fastq-filter and fastqdedup) where I run into this issue with more than 2 FASTQ files per readgroup. I still think it is nice to have dnaio solve this for me rather than adding custom code for every project.

> I think your most recent change of deprecating the file1 and file2 keyword arguments and instead just having a *files positional argument is a very nice improvement.

It is really weird. Sometimes programming for the general case is an excercise in pedantry which leads to byzantine and complex code, and in other cases it makes the API simpler. Bizarre. Is there a computer science theorem that tackles this problem? (I am a biologist by training).

I have committed al the nitpicks and simplified _open_paired. I also fixed an annoying mypy issue related to `sequence_class`.

marcelm:
> > Sorry about the very long delay. Are you still interested in merging this?
> 
> Yes. Aren't you? wink [marcelm/cutadapt#491 (comment)](https://github.com/marcelm/cutadapt/issues/491#issuecomment-1274754051) I have two applications (fastq-filter and fastqdedup) where I run into this issue with more than 2 FASTQ files per readgroup. I still think it is nice to have dnaio solve this for me rather than adding custom code for every project.

Absolutely, I’m just not sure I’ll be able to work on that issue in the foreseeable future. But this PR is the prerequisite for doing it at all.

> Bizarre. Is there a computer science theorem that tackles this problem? (I am a biologist by training).

I thought I heard something like this in a computer science context, but googling now only [gives me some results in mathematics](https://mathoverflow.net/questions/21214/). But this appears to be a thing, yes.

> I have committed al the nitpicks and simplified _open_paired. I also fixed an annoying mypy issue related to `sequence_class`.

I realized now that we need to make sure that the documentation covers the new classes as well. Do you want to do this here or in a separate PR? I can also help.

rhpvorderman:
> I thought I heard something like this in a computer science context, but googling now only [gives me some results in mathematics](https://mathoverflow.net/questions/21214/). But this appears to be a thing, yes.

Interesting. Thanks for the link.

> I realized now that we need to make sure that the documentation covers the new classes as well. Do you want to do this here or in a separate PR? I can also help.

I have added it just now. What should I do with `MultipleFileReader`? It is an interface obviously, but also, there is only one class that uses the interface. So I now have it documented as a plain class. That seemed to be the least convoluted way to solve this.

marcelm:
> What should I do with MultipleFileReader? It is an interface obviously, but also, there is only one class that uses the interface. So I now have it documented as a plain class. That seemed to be the least convoluted way to solve this.

I don’t have an opinion at the moment, let’s try it this way.

I think this is ready, let’s merge it!

marcelm:
I changed a setting on RTD to let it build the documentation on PRs. And I invited you as a co-maintainer.

rhpvorderman:
> I changed a setting on RTD to let it build the documentation on PRs. And I invited you as a co-maintainer.

Nice, that should keep the docstrings tidy. I accepted the invitation, thanks!

PR #88, opened 2022-08-02
Consistently raise FastaFormatError on parse errors

marcelm:
Raise it when an attempt is made to create a SequenceRecord with non-ASCII characters in one of the fields.

This is backwards compatible because `IsNotAscii` is a subclass of `ValueError`, which was previously raised for this type of problem.

Also, catch the exception in the FastaReader and convert it to a FastaFormatError when it occurs.

This will help in solving https://github.com/marcelm/cutadapt/issues/628.

### DIFF
--- a/src/dnaio/exceptions.py
+++ b/src/dnaio/exceptions.py
@@ -43,3 +43,28 @@ class FastaFormatError(FileFormatError):
     """
 
     format = "FASTA"
+
+
+class IsNotAscii(ValueError):
+    """
+    Some part of a FASTA or FASTQ file is not ASCII encoded
+
+    Attributes:
+        field: Description of the field ("name", "sequence", "qualities" or similar)
+            in which non-ASCII characters were found
+        value: Unicode string that was intended to be assigned to the field
+    """
+
+    def __init__(self, field: str, value: str):
+        self.value = value
+        self.field = field
+
+    def __str__(self):
+        detail = ""
+        try:
+            self.value.encode("ascii")
+        except UnicodeEncodeError as e:
+            detail = (
+                f", but found '{self.value[e.start:e.end]}' at position {e.start + 1}"
+            )
+        return f"'{self.field}' in sequence file must be ASCII encoded{detail}"
--- a/src/dnaio/exceptions.py
+++ b/src/dnaio/exceptions.py
@@ -43,28 +43,3 @@ class FastaFormatError(FileFormatError):
     """
 
     format = "FASTA"
-
-
-class IsNotAscii(ValueError):
-    """
-    Some part of a FASTA or FASTQ file is not ASCII encoded
-
-    Attributes:
-        field: Description of the field ("name", "sequence", "qualities" or similar)
-            in which non-ASCII characters were found
-        value: Unicode string that was intended to be assigned to the field
-    """
-
-    def __init__(self, field: str, value: str):
-        self.value = value
-        self.field = field
-
-    def __str__(self):
-        detail = ""
-        try:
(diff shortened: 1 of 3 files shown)


src/dnaio/exceptions.py
```diff
@@ -43,3 +43,28 @@ class FastaFormatError(FileFormatError):
...
+        self.value = value
+        self.field = field
+
+    def __str__(self):
+        detail = ""
+        try:
+            self.value.encode("ascii")
+        except UnicodeEncodeError as e:
+            detail = (
+                f", but found '{self.value[e.start:e.end]}' at position {e.start + 1}"
```
rhpvorderman, line 68: I don't know if the error is confusing to normal users, but it would be confusing to me (a slightly less than normal user). 
I would load the string and type: `my_string[position]` and that gives back the wrong character. Because this position is one-based. Can it be renamed to `at index {e.start}`?

src/dnaio/exceptions.py
```diff
@@ -43,3 +43,28 @@ class FastaFormatError(FileFormatError):
     """
 
     format = "FASTA"
+
+
+class IsNotAscii(ValueError):
```
rhpvorderman, line 48: I think the subclassing of the error does not add much unless there is a benefit to catching only errors of this class. 
In this case I think the ValueError is more informative. There is something wrong with a supplied value. The ASCII error part is already stated in the message.
Wouldn't a function that throws a ValueError with the same error message be a better option here?

marcelm: Yeah, I’ve pondered various options and maybe haven’t settled on the best one.

In the end, what I want to achieve is that `FastaReader.__iter__` throws a `FastaFormatError` whenever there’s a problem in the FASTA file. At the moment, it’s inconsistent because some parsing problems lead to `FastaFormatError`, but some to a `ValueError` (they bubble up from `SequenceRecord.__init__`). This is why I made the change below where I wrap `yield SequenceRecord...` in a `try: ... except`.

Since `IsNotAscii` is caught and immediately converted to a `FastaFormatError`, it may not make much sense to have an extra exception class just for this. I’ll change this.

src/dnaio/exceptions.py
marcelm: Hm, the idea is that the text can be printed unchanged when a parse error occurs. Cutadapt does this, for example:
```
cutadapt: error: Error in FASTA file at line 5: 'name' in sequence file must be ASCII encoded, but found '☺' at position 1
```
It’s also consistent with the 1-based line number in `FileFormatError.__str__`: https://github.com/marcelm/dnaio/blob/095c08cf1fc7957a9e739342a4aef03344cc874e/src/dnaio/exceptions.py#L28

I see what you mean, though. IMO, enduser-visible numbers should be 1-based, and developers should get 0-based ones, but the exception message is shown both to users and developers, so there’s a conflict. IDK, maybe directly showing `str()` representations of exceptions to the user isn’t the best architecture (but it is pragmatic).

src/dnaio/exceptions.py
marcelm: Another idea was to just throw the `UnicodeEncodeError` (it’s a subclass of ValueError), but that loses the information *which* of the values wasn’t ASCII (name/sequence/qualities).

rhpvorderman (approved):
Looks good! My feedback was mostly inspired by https://youtu.be/o9pEzgHorH0?t=617

marcelm:
Yes, good talk, I should re-watch it once a year so I don’t forget :-)!

PR #91, opened 2022-10-07
Fix mypy error: A function returning TypeVar should receive at least …

marcelm:
…one argument containing the same Typevar


rhpvorderman:
Now I start wondering if we just can't remove the whole sequence_class to fix this issue ;-).

PR #96, opened 2022-10-26
Allow setting open_threads and compression_level

### LINKED ISSUE #90: compresslevel and open_threads should be arguments in the dnaio.open interface
Hi, I have been working on the benchmarking and I am quite intrigued by fastp. The project is very similar to dnaio in terms of it API. Paired reading, writing, a focus on speed, using zlib-compatible libraries to speed up (de)compression etc.

Anyway, I noticed that the fastq.gz reading benchmark between dnaio and fastp is not entirely fair, as the benchmark program always uses a single thread and dnaio gets around the GIL by launching an `igzip` process. This made me realize that in al my programs I do this:
```python
import functools
import xopen

DEFAULT_OPENER = functools.partial(xopen, threads=0, compresslevel=1))

# Use all dnaio.open calls with DEFAULT_OPENER
```
This is quite an ugly way to interact with an API. 

I propose that we add `open_threads=0` and `compresslevel=1` to the `dnaio.open` interface. `open_threads` should be 0, as any polite program will only go multithreaded when explicitly asked (unless it has parallel in its name). This will use python-isal by default, which is more efficient than piping through igzip (in terms of total cpu time, not wall clock time).  We can add a nice docstring to the parameter that explains an external program will be opened when threads > 0.  

Compresslevel should be set at 1. This is the most efficient compression level in terms of compute time for a given final size. Tools producing FASTQ almost always produce intermediate files (they are mapped afterwards), so the most common use case will be one where throughput is most important. An advantage of having a `compresslevel` parameter in `dnaio.open` is that it is very easy for users to change the parameter if they feel compression is more important than throughput. I know you have argued differently for cutadapt, but IIRC that was primarily motivated by backwards-compatibility concerns, where changing the behaviour of the program drastically was undesirable (which I agree with). Dnaio is not as well established, and in fact, still in its 0.x days. 

I would have proposed this change with a PR, but since #87 also messes with the open function I leave this for after that is merged (or rejected).

rhpvorderman:
A very minor addition to the open function, but making it much more convenient. Compatibility with cutadapt is kept.

This fixes #90 . 

Discussion points:
- compresslevel vs compression_level. I think that compresslevel is not a very good abbreviation, on the other hand it is used consistently throughout CPython (and also xopen). Since dnaio is not a compression library I think the non-abbreviated version is better here.
- open_threads 0 by default. As someone who regularly works on a compute cluster I am regularly annoyed by applications that by default use all threads or a lot of them. One reason is that when I start a simple script/program on the login node this shouldn't immediately drain all the CPUs if I haven't told it so. This makes the login node unusable, so in my mind the only sane default is 1 thread, unless "parallel" is in the name of the application. (Pigz thus gets a free pass for wanting to use all threads). The other reason is that extra threads come with extra overhead. Since I usually run several samples at the same time I rather use less total compute time as I can run everything in parallel anyway to drive the wall clock time down. 
 
My opinions on the above are not absolute. As long as the open_threads parameter goes in I can choose a default for my own applications anyway. So I wonder what you think. As some extra information here are the benchmarks:

Running with open_threads 0. Total compute time 2.571 + 0.054 = 2.625
```
Benchmark 1: python dnaio_read.py ~/test/ERR039854.fastq.gz
  Time (mean ± σ):      2.625 s ±  0.085 s    [User: 2.571 s, System: 0.054 s]
  Range (min … max):    2.516 s …  2.748 s    10 runs
``` 

Running with open_threads=1. Decompress process is igzip version 2.30 (the fastest option). Total compute time 2.391 + 0.454 = 2.845
```
Benchmark 1: python dnaio_read.py ~/test/ERR039854.fastq.gz
  Time (mean ± σ):      1.560 s ±  0.035 s    [User: 2.391 s, System: 0.454 s]
  Range (min … max):    1.521 s …  1.626 s    10 runs
 ```

For comparison: reading the uncompressed file. The system time goes up a bit compared to 0 threaded gzip open because the filesystem is queried less on the smaller gzip file.
```
Benchmark 1: python dnaio_read.py ~/test/ERR039854.fastq
  Time (mean ± σ):      1.131 s ±  0.024 s    [User: 1.000 s, System: 0.131 s]
  Range (min … max):    1.107 s …  1.184 s    10 runs
```
 
My conclusion is that there is quite a penalty to be payed for using an external proess in terms of total compute time. This wasn't the case when I had just made python-isal, but by this point it is so optimized that it is much better than piping trough igzip.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,11 @@ vnext
 * :pr:`99`: SequenceRecord initialization is now faster which also provides
   a speed boost to FASTQ iteration. ``SequenceRecord.__new__`` cannot be used
   anymore to initialize `SequenceRecord` objects.
+* :pr:`96`: ``open_threads`` and ``compression_level`` are now added
+  to `~dnaio.open` as arguments. By default dnaio now uses compression level
+  1 and does not utilize external programs to speed up gzip (de)compression.
+* :pr:`87`: `~dnaio.open` can now als open more than two files.
+  The ``file1`` and ``file2`` arguments are now deprecated.
 
 v0.9.1 (2022-08-01)
 -------------------
--- a/src/dnaio/__init__.py
+++ b/src/dnaio/__init__.py
@@ -30,6 +30,7 @@ __all__ = [
     "__version__",
 ]
 
+import functools
 from os import PathLike
 from typing import Optional, Union, BinaryIO
 
@@ -85,7 +86,9 @@ def open(
     interleaved: bool = False,
     mode: str = "r",
     qualities: Optional[bool] = None,
-    opener=xopen
+    opener=xopen,
+    compression_level: int = 1,
+    open_threads: int = 0,
 ) -> Union[
     SingleEndReader,
     PairedEndReader,
@@ -136,6 +139,17 @@ def open(
         already open file-like objects. By default, ``xopen`` is used, which can
         also open compressed file formats.
 
+      open_threads: By default dnaio opens files in the main thread,
+        when threads is greater than 0 external processes are opened for
+        compressing and decompressing files. This decreases wall clock time
+        at the cost of a little extra overhead. This parameter does not work
+        when a custom opener is set.
+
+      compression_level: By default dnaio uses compression level 1 for writing
+        gzipped files as this is the fastest. A higher level can be set using
+        this parameter. This parameter does not work when a custom opener is
+        set.
+
     Return:
        A subclass of `SingleEndReader`, `PairedEndReader`, `SingleEndWriter` or
        `PairedEndWriter`.
@@ -165,6 +179,10 @@ def open(
     elif mode not in ("r", "w", "a"):
         raise ValueError("Mode must be 'r', 'w' or 'a'")
 
+    if opener == xopen:
+        opener = functools.partial(
+            xopen, threads=open_threads, compresslevel=compression_level
+        )
     if interleaved or len(files) == 2:
         return _open_paired(
             *files,
--- a/src/dnaio/__init__.py
+++ b/src/dnaio/__init__.py
@@ -116,8 +116,8 @@ def open(
         Set to ``'r'`` for reading, ``'w'`` for writing or ``'a'`` for appending.
 
       interleaved:
-        If True, then file1 contains interleaved paired-end data.
-        file2 must be None in this case.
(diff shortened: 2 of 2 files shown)


marcelm:
I’m inclined to agree with the `open_threads=0` default, but give me a couple of days to think about this, and also the other change.

marcelm:
Can you clarify: The first benchmark result is for *not* for `open_threads=0`, I assume?

rhpvorderman:
Whoops an oversight in my text I see. I have two 0's. Now it correctly states that the first test is open_threads=0 and the second test is open_threads=1. 
As you can see, open_threads=0 leads to a longer wall-clock time, but a shorter compute time. open_threads=1 has a shorter wall-clock time, but pays for the overhead of the pipe and the threads with extra compute time.

marcelm (approved):
So the reason I’ve not gotten back to this for so long is that I thought that this would also change the default compression level to 1. However, it appears that this is already implicitly the case when using any xopen version >=1.0.0 where we switched to using an external `igzip` process by default because `igzip` has default compression level 1. Sorry for the delay.

Can you also add a changelog entry? Then we can merge this and release 0.10.0 as you suggested.

rhpvorderman:
> So the reason I’ve not gotten back to this for so long is that I thought that this would also change the default compression level to 1. However, it appears that this is already implicitly the case when using any xopen version >=1.0.0 where we switched to using an external igzip process by default because igzip has default compression level 1. Sorry for the delay.

No problem. I think this change will do much good now that the compression is transparantly and explicitly set to 1 always. That will lead to less unpredictable behaviour. Same goes for setting compression threads to 0.

Updated the changelog entry also with the changes from the multiple API. As well as fixing a typing error that mypy caught.

marcelm:
> I think this change will do much good now that the compression is transparantly and explicitly set to 1 always. That will lead to less unpredictable behaviour. Same goes for setting compression threads to 0.

Yes, it was a bit fuzzy before; this is an improvement.

> Updated the changelog entry also with the changes from the multiple API. As well as fixing a typing error that mypy caught.

Thanks.

> EDIT: Also did a double check on the latest cutadapt main to see if this API change does not break cutadapt. No test failures, so it should be fine. (Famous last words...)

I’ll merge this and then also run a check myself, just to be sure.

PR #99, opened 2022-10-28
Faster SequenceRecord initialization.

rhpvorderman:
Before:

```
Benchmark 1: python dnaio_read.py ~/test/big2.fastq
  Time (mean ± σ):     914.5 ms ±  23.8 ms    [User: 792.7 ms, System: 121.7 ms]
  Range (min … max):   882.0 ms … 950.4 ms    10 runs
```

After:
```
Benchmark 1: python dnaio_read.py ~/test/big2.fastq
  Time (mean ± σ):     752.6 ms ±  14.0 ms    [User: 628.9 ms, System: 123.5 ms]
  Range (min … max):   725.8 ms … 779.4 ms    10 runs
```

The disadvantage is that we break `__new__` for everyone. Is this a problem for cutadapt? `__init__` works fine.

The speed advantage requires some explanation. Using the CPython API there are two methods you can set on your extension class: `tp_new` and `tp_init`. `tp_new` creates the class and `tp_init` initializes it. Most extension classes have however, no dynamic attributes so do not need further initialization. tp_new also accepts arguments and keyword arguments and can therefore be used to set the basic attributes. Therefore some objects such as the python `bytes` object only use the `tp_new` slot.
Also in the Python C-API it is possible to forego the `tp_new` slot entirely and simply initialize objects using `PyObject_New` and set struct members manually.  This is not possible in Cython as Cython needs full control over how objects are initialized (it might add/need extra members to the struct, so it is not possible to preduct this beforehand).

I did some digging and this is how the tp_new function looked before:
```C
static PyObject *__pyx_tp_new_5dnaio_5_core_SequenceRecord(PyTypeObject *t, PyObject *a, PyObject *k) {
  struct __pyx_obj_5dnaio_5_core_SequenceRecord *p;
  PyObject *o;
  if (likely((t->tp_flags & Py_TPFLAGS_IS_ABSTRACT) == 0)) {
    o = (*t->tp_alloc)(t, 0);
  } else {
    o = (PyObject *) PyBaseObject_Type.tp_new(t, __pyx_empty_tuple, 0);
  }
  if (unlikely(!o)) return 0;
  p = ((struct __pyx_obj_5dnaio_5_core_SequenceRecord *)o);
  p->_name = Py_None; Py_INCREF(Py_None);
  p->_sequence = Py_None; Py_INCREF(Py_None);
  p->_qualities = Py_None; Py_INCREF(Py_None);
  if (unlikely(__pyx_pw_5dnaio_5_core_14SequenceRecord_1__cinit__(o, a, k) < 0)) goto bad;
  return o;
  bad:
  Py_DECREF(o); o = 0;
  return NULL;
}
```
As you can see, it always calls the `__cinit__` function which has uses arguments (`a`) and keyword arguments (`k`). Most of what `__cinit__` does is parsing the argument tuple and keyword dict. This is expensive. When `__cinit__` is removed this is the tp_new function.

```C
static PyObject *__pyx_tp_new_5dnaio_5_core_SequenceRecord(PyTypeObject *t, CYTHON_UNUSED PyObject *a, CYTHON_UNUSED PyObject *k) {
  struct __pyx_obj_5dnaio_5_core_SequenceRecord *p;
  PyObject *o;
  if (likely((t->tp_flags & Py_TPFLAGS_IS_ABSTRACT) == 0)) {
    o = (*t->tp_alloc)(t, 0);
  } else {
    o = (PyObject *) PyBaseObject_Type.tp_new(t, __pyx_empty_tuple, 0);
  }
  if (unlikely(!o)) return 0;
  p = ((struct __pyx_obj_5dnaio_5_core_SequenceRecord *)o);
  p->_name = Py_None; Py_INCREF(Py_None);
  p->_sequence = Py_None; Py_INCREF(Py_None);
  p->_qualities = Py_None; Py_INCREF(Py_None);
  return o;
}
```
As you can see, it adds the UNUSED macros to the arguments and keyword arguments. By default it sets everything to `None` which is a waste of resources, but not as much as parsing keyword arguments.

TLDR: I did it! Finally I, managed to make Cython do the most barebones init possible, while still letting Cython do the init and not relying on hacks trough the C-API. This should be stable regardless of any new Cython version going forward.

EDIT: I kept a branch in pure C around. It was very useful for learning some parts of the C-API and asses the maximum capability of our current approach. It performs like this:
```
Benchmark 1: python dnaio_read.py ~/test/big2.fastq
  Time (mean ± σ):     683.2 ms ±  19.7 ms    [User: 565.2 ms, System: 117.7 ms]
  Range (min … max):   653.1 ms … 710.8 ms    10 runs
```
So it is even faster (~10%) because it foregoes all the Cython stuff, but this comes at a cost of much verbose. We have already discussed this and dumping Cython which you are very familiar with, for a verbose C code base that is more difficult to maintain is not helpful. Having the speed difference down to 10% makes it definitely not worth the switch. (Still I prefer C for my own projects as getting Cython to do the things I actually want has been quite an annoying and time-consuming process ;-). But finally, that is now finished for dnaio.)

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -85,14 +85,7 @@ cdef class SequenceRecord:
         object _sequence
         object _qualities
 
-    def __cinit__(self, object name, object sequence, object qualities=None):
-        """Set qualities to None if there are no quality values"""
-        self._name = name
-        self._sequence = sequence
-        self._qualities = qualities
-
     def __init__(self, object name, object sequence, object qualities = None):
-        # __cinit__ is called first and sets all the variables.
         if not PyUnicode_CheckExact(name):
             raise TypeError(f"name should be of type str, got {type(name)}")
         if not PyUnicode_IS_COMPACT_ASCII(name):
@@ -111,6 +104,9 @@ cdef class SequenceRecord:
                 raise ValueError("In read named {!r}: length of quality sequence "
                                  "({}) and length of read ({}) do not match".format(
                     rname, len(qualities), len(sequence)))
+        self._name = name
+        self._sequence = sequence
+        self._qualities = qualities
 
     @property
     def name(self):
@@ -299,6 +295,7 @@ cdef class SequenceRecord:
             char *qualities
             Py_ssize_t cursor, reverse_cursor
             unsigned char nucleotide
+            SequenceRecord seq_record
         reverse_cursor = sequence_length
         for cursor in range(sequence_length):
             reverse_cursor -= 1
@@ -315,8 +312,11 @@ cdef class SequenceRecord:
                 reversed_qualities[reverse_cursor] = qualities[cursor]
         else:
             reversed_qualities_obj = None
-        return SequenceRecord.__new__(
-            SequenceRecord, self._name, reversed_sequence_obj, reversed_qualities_obj)
+        seq_record = SequenceRecord.__new__(SequenceRecord)
+        seq_record._name = self._name
+        seq_record._sequence = reversed_sequence_obj
+        seq_record._qualities = reversed_qualities_obj
+        return seq_record
 
 
 def paired_fastq_heads(buf1, buf2, Py_ssize_t end1, Py_ssize_t end2):
@@ -490,6 +490,7 @@ cdef class FastqIter:
     def __next__(self):
         cdef:
             object ret_val
+            SequenceRecord seq_record
             char *name_start
             char *name_end
             char *sequence_start
@@ -593,8 +594,11 @@ cdef class FastqIter:
             if self.use_custom_class:
                 ret_val = self.sequence_class(name, sequence, qualities)
             else:
-                ret_val = SequenceRecord.__new__(SequenceRecord, name, sequence, qualities)
-
+                seq_record = SequenceRecord.__new__(SequenceRecord)
+                seq_record._name = name
+                seq_record._sequence = sequence
+                seq_record._qualities = qualities
+                ret_val = seq_record
             # Advance record to next position
             self.number_of_records += 1
             self.record_start = qualities_end + 1
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+vnext
+-------------------
+* :pr:`99`: SequenceRecord initialization is now faster which also provides
+  a speed boost to FASTQ iteration. ``SequenceRecord.__new__`` cannot be used
+  anymore to initialize `SequenceRecord` objects.
+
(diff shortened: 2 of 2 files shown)


rhpvorderman:
The post above is quite a ramble. Let me summarize now that I have a more clear head:
Any python object needs to be allocated in memory. This is the purpose of the `__new__` function. The `__new__` function can also be used to initialize the object if its attributes are static.
The `__init__` function is used on objects on which its attributes are dynamic and context dependent. Python automatically calls the objects `__new__` function before `__init__` is called. 

Cython implements it as follows.
- The `__new__` function is called to initialize a cdef class. In this `__new__` function all python object members are set to None.
- The `__cinit__` function is called from the `__new__` function to initialize all the C members of the object (if the user has specified `__cinit__`). `__cinit__` does not take C arguments, but an a python argument tuple and keyword dict instead. Which are parsed.
- The `__init__` function is implemented by the user. This also takes a python argument tuple and keyword dict. 

By leaving out `__cinit__` entirely, it is not called anymore. That means that an extra step parsing an argument tuple and a keyword dict is left out. This is a major source of slowdown. We don't need it because there are no C attributes that need initializing in the SequenceRecord class.

src/dnaio/_core.pyx
```diff
@@ -85,12 +85,6 @@ cdef class SequenceRecord:
...
         object _qualities
 
-    def __cinit__(self, object name, object sequence, object qualities=None):
-        """Set qualities to None if there are no quality values"""
-        self._name = name
-        self._sequence = sequence
-        self._qualities = qualities
-
     def __init__(self, object name, object sequence, object qualities = None):
         # __cinit__ is called first and sets all the variables.
```
marcelm, line 89: This can be deleted now

marcelm (approved):
Cool! Thanks also for the second comment; that was helpful. Cutadapt doesn’t use `__new__`, so that not being available isn’t a problem, and I think `__init__` is indeed good enough.

Can you add a changelog entry? Then feel free to merge!

Probably time for a release (0.9.2) afterwards. I’ll look into that.

rhpvorderman:
>  and I think __init__ is indeed good enough.

By not calling `__cinit__` (via `__new__`), `__init__` should also be faster. Unfortunately this is hard to verify with timeit. The old `__cinit__` code was pages long, mostly for parsing arguments (it hardly did anything else). I am glad we have been able to optimize it out.

rhpvorderman:
> Probably time for a release (0.9.2) afterwards. I’ll look into that.

I propose merging #96 too and bumping to 0.10.0. Then we can use that version for the benchmarks in the paper as it should be faster.

PR #100, opened 2022-11-01
Let’s ignore __repr__ functions for coverage percentages

marcelm:


rhpvorderman:
`__repr__` is only used on the repl and for debugging. Adding tests feels a bit pedantic indeed. On the other hand, having a fatal error in a repr function can really ruin your day. Also coverage is just a metric, there is no reason to inflate it.

I can add a few `__repr__` tests and then we can have the best of both worlds.

PR #101, opened 2022-11-01
Avoid running CI jobs twice

marcelm:
Jobs are triggered twice if they are pushed to the repo and if they are part of a pull request. With this, a job belonging to a pull request is only run if it comes from a forked repository.

Derived from
https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/, (but typos fixed)

(It would be nicer if GitHub just handled this automatically.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,10 @@ on: [push, pull_request]
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


marcelm:
Seems to be working: The pull_request jobs are marked as "Skipped".

PR #105, opened 2022-11-15
Check for most common case where there is no second header

rhpvorderman:
This is a very minor speed improvement, but it is measurable. Especially on reads with only 43bp, where records are small and the +lines make up relatively more of the file. The difference is still measurable on 150bp given a reasonably large file. On 9000 bp it does not matter.

before:
```
Benchmark 1: python dnaio_read.py ../dnaio-benchmarks/data/single43bp.fastq
  Time (mean ± σ):      1.240 s ±  0.034 s    [User: 1.154 s, System: 0.086 s]
  Range (min … max):    1.199 s …  1.307 s    10 runs
```

After
```
Benchmark 1: python dnaio_read.py ../dnaio-benchmarks/data/single43bp.fastq
  Time (mean ± σ):      1.179 s ±  0.035 s    [User: 1.084 s, System: 0.095 s]
  Range (min … max):    1.128 s …  1.258 s    10 runs
```

I also tested it on my `allinc` branch. The difference is smaller there, but since the C branch has smaller run-to-run variance it is still clearly visible. 

Can you reproduce this on your system? If so, it might be nice to put it in before 0.10.0 to have it count for the benchmarks in the paper. 

For other performance improvements I checked whether I could improve cache locality by rearranging the pointer dereferences. The code is now ordered nicely by function, which makes it more readable, but not by memory access. So ideally you want to check for `@` first, then search the newline, then check for `\r` before the newline etc. That way you don't perform jumps. So I did a small test, but it does not matter. Even not on the 9000bp file. Maybe I improved cache locality, but we keep a max of 1 record in cache. And L1 cache latency is 3-4 clockticks. So that is why it does not matter probably. So it is better to keep it this way, where the code is much more readable.

I also want to see if we can use more knowledge of the format to parse even faster. I think the `+\n` case is a good example. Another case is the qualities line. The qualities line should be the same length as the sequence line. Still we scan for `\n`. But we have to do that, because when files are concatenated (quite common) it might occur that a truncated file concatenated to another file yields a valid result if we do not do the check properly. Still we do know where the `\n` should be and I am hoping that we find a way to utilize that knowledge to make it faster. It will be hard, because `memchr` is basically hand-written assembler and it uses sse2, avx2 or avx512 dependening on a runtime check on your system. It checks one cacheline (64B or 512bits) at the same time. So it is incredibly hard to beat.

### DIFF
--- a/src/dnaio/_core.pyx
+++ b/src/dnaio/_core.pyx
@@ -507,6 +507,7 @@ cdef class FastqIter:
             char *qualities_start
             char *qualities_end
             char *buffer_end
+            size_t remaining_bytes
             Py_ssize_t name_length, sequence_length, second_header_length, qualities_length
         # Repeatedly attempt to parse the buffer until we have found a full record.
         # If an attempt fails, we read more data before retrying.
@@ -532,10 +533,15 @@ cdef class FastqIter:
                 self._read_into_buffer()
                 continue
             second_header_start = sequence_end + 1
-            second_header_end = <char *>memchr(second_header_start, b'\n', <size_t>(buffer_end - second_header_start))
-            if second_header_end == NULL:
-                self._read_into_buffer()
-                continue
+            remaining_bytes = (buffer_end - second_header_start)
+            # Usually there is no second header, so we skip the memchr call.
+            if remaining_bytes > 2 and second_header_start[0] == b'+' and second_header_start[1] == b'\n':
+                second_header_end = second_header_start + 1
+            else:
+                second_header_end = <char *>memchr(second_header_start, b'\n', <size_t>(remaining_bytes))
+                if second_header_end == NULL:
+                    self._read_into_buffer()
+                    continue
             qualities_start = second_header_end + 1
             qualities_end = <char *>memchr(qualities_start, b'\n', <size_t>(buffer_end - qualities_start))
             if qualities_end == NULL:


marcelm:
Thanks, I’l check this in a couple of days, busy with a lecture at the moment.

rhpvorderman:
I did a check if I could do something similar to #14 on a `uselength` branch to improve quality line speed, but I have to conclude that even if I did something safely at 0 cost (which is impossible) the gains would be extremely small. On the same magnitude as this PR. Which is not great. Even when reads have 9000 phred scores each, which are all scanned for '\n'. The scanning is so fast, that the actual overhead of dereferencing the `memchr` function pointer accounts for most of the cost. (As shown in this PR, there is a cost to calling memchr as opposed to not doing it). So I think by this point it is almost impossible to improve the parsing code further and most of the gains can be produced by reducing the C to Python overhead.

marcelm:
I’ve tried to measure this, but don’t have a more or less unused system available for this at the moment, so the variance is too high and I couldn’t see a difference. I’ll just merge this.

Looking at your efforts, I wonder if there’s some great idea out there that would improve this even further significantly ...

Ok to release 0.10.0?

rhpvorderman:
> so the variance is too high and I couldn’t see a difference.

It really only helps quite short reads. But at least it does not make it slower in any of the other cases.

> Ok to release 0.10.0?

Yes! I am looking forward to it! Especially with all the updates to the open function.

Another note: I have been quite busy lately, but I am planning to work on the dnaio graphs for the paper during the holiday.

marcelm:
Ok, I’ve made the release. Nice round number :smile:.

> I have been quite busy lately, but I am planning to work on the dnaio graphs for the paper during the holiday.

Same here, I hope I can set aside a couple of hours.

PR #106, opened 2022-12-13
Fix buffer filling up on incomplete paired FASTQ

marcelm:
When fed paired-end FASTQ where one of the files contains more records than the other, read_paired_chunks() now detects that situation and raises an error instead of filling up the buffer. Filling up the buffer also leads to an error, but a misleading one.

See https://github.com/marcelm/cutadapt/issues/660


rhpvorderman:
Good catch, it was quite a puzzling mystery.

PR #110, opened 2023-08-30
Add id and comment

### LINKED ISSUE #10: Add Sequence.id attribute
and also `Sequence.comment` and perhaps `Sequence.sep`. They could be lazily computed. Perhaps also add `.header` as an alias for `.name`? Should setting `.id` and `.comment` be supported? Convenient, but perhaps slower than setting the full header directly.

rhpvorderman:
You had something like this in mind?

EDIT:
fixes #10


src/dnaio/_core.pyx
```diff
@@ -150,6 +154,47 @@ cdef class SequenceRecord:
...
+        # Not yet cached is None
+        if self._comment is None:
+            name = <char *>PyUnicode_DATA(self._name)
+            name_length = <size_t>PyUnicode_GET_LENGTH(self._name)
+            id_length = strcspn(name, "\t ")
+            if id_length == name_length:
+                self._comment = ""
+            else:
+                comment_length = name_length - (id_length + 1)
+                comment_start = name + id_length + 1
```
marcelm, line 190: Hm, this assumes that there’s only one whitespace between the id and the comment. I’d assume this is true most of the time, but I think we need to be a bit more defensive.

src/dnaio/_core.pyx
```diff
@@ -150,6 +154,47 @@ cdef class SequenceRecord:
...
+                self._comment = PyUnicode_New(comment_length , 127)
+                memcpy(PyUnicode_DATA(self._comment), comment_start, comment_length)
+        # Cached but nothing is internally empty string, expose externally as None
```
marcelm, line 193: I don’t get the comment. Maybe just remove? Would be ok

tests/test_records.py
```diff
@@ -105,6 +105,50 @@ def test_set_qualities_none(self):
...
+            seq.comment = "Hello there!"
+
+    @pytest.mark.parametrize(
+        ["record", "expected"],
+        [
+            (SequenceRecord("name", "A", "="), None),
+            (SequenceRecord("name ", "A", "="), None),
+            (SequenceRecord("name  ", "A", "="), " "),
+            (SequenceRecord("name", "A", "="), None),
+            (SequenceRecord("AotC I hate sand!", "A", "="), "I hate sand!"),
```
marcelm, line 125: :smile:

marcelm (commented):
Yes, very nice! This will be really nice to have. We should make a release as soon as this is merged.

This also needs a changelog and documentation update. (I can do that if you don’t have time over the coming days.)

src/dnaio/_core.pyx
rhpvorderman: Finally, a use for strspn!

src/dnaio/_core.pyx
rhpvorderman: I will try to make it more clear in the comment. It is a rather odd design choice so it needs one, I think.

tests/test_records.py
rhpvorderman: https://www.youtube.com/watch?v=C_yWiW7OxZw

rhpvorderman:
All done!

tests/test_records.py
marcelm: Thanks, I got the Star Wars reference, but didn’t know the song (and know it’ll be stuck in my head for the rest of the day ...)

rhpvorderman: Well you could always listen to "Last jedi knight" by the same musician to get it out of your head again ;-).

PR #112, opened 2023-08-31
Get rid of setup.cfg, use pyproject.toml instead

marcelm:
This modernizes our project metadata configuration and allows us to get rid of `setup.cfg` entirely.


rhpvorderman (approved):
Nice!

PR #117, opened 2023-10-04
Bump various versions

marcelm:
In particular, drop Python 3.7 and add Python 3.12.


marcelm:
Shouldn’t be controversial, so merging this now.

Issue #94, opened 2022-10-24
Add better type hints for dnaio.open()


marcelm:
with overloads, similar to https://github.com/pycompression/xopen/pull/112

PR #120, opened 2023-10-05
Improve type hints

marcelm:
* Run mypy on `tests/`
* Type-annotate the tests (add a `-> None` to most of the functions) so that mypy checks them
* Add overloads for `open()` as suggested in #94. To make mypy happy, I needed to add a `_kwargs` argument to `open()`.
* Add configuration to make mypy a little bit stricter (and remove the unused `type: ignore` comments that this found)

CC @fjossandon


rhpvorderman (approved):
That is a non-trivial amount of work. Great!

marcelm:
Thanks! I always have to read up on how this is done (in particular overloads) because I do this so rarely.

I also added `-> None` to the test functions in `test_open.py` to get them typechecked as well. This required me to drop the deprecated `file1` and `file2` arguments from the overloads and from those places where they are still used in the tests. I spent way too much time on trying to make it work, but it was easier to just do it this way.

I’ll look into the coverage issue in a separate PR.

PR #122, opened 2023-10-27
Add asan tox environment for testing memory errors

rhpvorderman:
None were found. Unfortunately the environment cannot run on github CI. This might be useful for some of your other projects. I just managed to remove two memory errors from python-isal. They were in exception paths, nothing major, but still nice.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -38,7 +38,7 @@ setenv=
     CFLAGS=-lasan -fsanitize=address -fno-omit-frame-pointer
 allowlist_externals=bash
 commands=
-    bash -c 'export LD_PRELOAD=$(find /usr/lib -name "libasan.so*" | head -n 1) && printenv LD_PRELOAD && python -c "import dnaio" && pytest tests'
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -31,6 +31,15 @@ deps =
...
 commands = mypy src/ tests/
 
+[testenv:asan]
+setenv=
+    PYTHONDEVMODE=1
+    PYTHONMALLOC=malloc
+    CFLAGS=-lasan -fsanitize=address -fno-omit-frame-pointer
+allowlist_externals=bash
+commands=
+    bash -c 'export LD_PRELOAD=$(find /usr/lib -name "libasan.so*" | head -n 1) && printenv LD_PRELOAD && python -c "import dnaio" && pytest tests'
```
marcelm, line 41: This crashes for me because apparently the `find ... | head ...` appears to be fragile. I found [this SO answer](https://stackoverflow.com/a/47022141), which recommended using `gcc -print-file-name=libasan.so` instead. That worked for me.

rhpvorderman: Fixed!

PR #123, opened 2023-11-20
Highlight xopen's Zstandard support when optionally installed

bede:
Hi, what a fantastically useful library this is! I particularly like that it supports stdin and Zstandard compression.

This tiny PR highlights underlying xopen's (and therefore dnaio's) existing support for `.zst` extensions, mentioning that either of two dependencies (`zstd` CLI or `Zstandard` PyPI package) must be installed to make it work.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ Features and supported file types
 - FASTQ input and output
 - FASTA input and output
 - BAM input
-- Compressed input and output (``.gz``, ``.bz2`` and ``.xz``, detected automatically)
+- Compressed input and output (``.gz``, ``.bz2``, ``.xz`` and ``.zst`` are detected automatically; z)
 - Paired-end data in two files
 - Interleaved paired-end data in a single file
 - Files with DOS/Windows linebreaks can be read
@@ -50,6 +50,8 @@ Limitations
 
 - Multi-line FASTQ files are not supported.
 - FASTQ and BAM parsing is the focus of this library. The FASTA parser is not as optimized.
+- For reading and writing Zstandard (``.zst``) files, either the ``zstd`` command-line
+program or the Python ``zstandard`` package needs to be installed.
 
 Links
 =====
--- a/README.rst
+++ b/README.rst
@@ -50,8 +50,7 @@ Limitations
 
 - Multi-line FASTQ files are not supported.
 - FASTQ and BAM parsing is the focus of this library. The FASTA parser is not as optimized.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -39,7 +39,7 @@ Features and supported file types
 - FASTQ input and output
 - FASTA input and output
 - BAM input
-- Compressed input and output (``.gz``, ``.bz2`` and ``.xz``, detected automatically)
+- Compressed input and output (``.gz``, ``.bz2``, ``.xz`` and ``.zst`` are detected automatically; z)
```
marcelm, line 42: There’s a spurious `; z`

bede:
Indeed there was, apologies!

README.rst
```diff
@@ -50,6 +50,7 @@ Limitations
 
 - Multi-line FASTQ files are not supported.
 - FASTQ and BAM parsing is the focus of this library. The FASTA parser is not as optimized.
+- For reading and writing Zstandard (``.zst``) files, either the ``zstd`` command-line program or the Python ``zstandard`` package needs to be installed.
```
marcelm, line 53: Please add back the linebreak you had in an earlier comment. You just need to ensure you indent after the linebreak:

```
- First item
- Second
  item
- Third item
```

marcelm:
Thanks! This should definitely be mentioned in the docs. It should probably not be under "Limitations". The idea is that Zstandard support is optional: If you want it, you need to use `pip install dnaio zstandard`.

We delegate compression/decompression to https://github.com/pycompression/xopen/, which also has zstandard as an optional dependency. The reason for it being optional is that the zstandard wheels are quite large. See https://github.com/pycompression/xopen/pull/91#issuecomment-1287495260 (and the subsequent comment). The numbers have gotten even worse: The wheel is 5.2 MB and it needs 23 MB on disk unpacked. This could be 800 kB (2 MB unpacked) if debugging symbols were removed.

README.rst
bede: Thanks, I'm not very familiar with RST

marcelm:
How about adding an "Installation" section after "Example usage"? Just a short sentence mentioning `pip install dnaio zstandard` and say that zstandard can be omitted if support for `.zst` files is not needed.

And an addition to the "Compressed input and output" bullet point would also be good.

bede:
Thanks, I've added an installation section as suggested, thoughts?

bede:
And thanks for developing this great library!

PR #130, opened 2024-02-02
Run tests on the M1 macOS runner

marcelm:
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

We should also generate wheels at some point, but cibuildwheel only has preliminary support for it; we can wait a little bit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,9 @@ jobs:
         python-version: ["3.10"]
         toxenv: [flake8, black, mypy, docs]
     steps:
-    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Very nice! All those poor people who got duped into buying one because "you can also use Macs for development" only to find out their proprietary architecture had hardly any native software support in the data science space... Oh well. It will take some time for the entire stack to be ready. ISA-L does not have M1 mac support yet (in the works). Zlib-ng does, but conda-forge is not yet compiling that package for M1 natively. I don't know about bioconda, but with a native runner that should catch up quickly.

PR #131, opened 2024-04-03
Use dispatch rather than rely on separate compile commands for SSSE3 bam parser

rhpvorderman:
See: https://github.com/samtools/htslib/pull/1764

Using dispatch is less flaky than precompiling with SSSE3 instructions. It will work on any machine.

I made the dispatching work properly with GCC. I also used the basic nibble2base function in htslib as a base for the default. 

From prior discussion in htslib I learned that deoding the qualities can be easily optimized by the compiler including the tricks to do so.

### DIFF
--- a/src/dnaio/bam.h
+++ b/src/dnaio/bam.h
@@ -6,14 +6,21 @@
 # define GCC_AT_LEAST(major, minor) 0
 #endif
 
-#ifdef __clang__
-#ifdef __has_attribute
+#if defined(__clang__) && defined(__has_attribute)
 #define CLANG_COMPILER_HAS(attribute) __has_attribute(attribute)
-#endif 
-#endif
-#ifndef CLANG_COMPILER_HAS 
+#else
 #define CLANG_COMPILER_HAS(attribute) 0
-#endif 
+#endif
+
+#define COMPILER_HAS_TARGET (GCC_AT_LEAST(4, 8) || CLANG_COMPILER_HAS(__target__))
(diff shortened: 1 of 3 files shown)


src/dnaio/bam.h
```diff
@@ -1,21 +1,26 @@
...
+#define GCC_AT_LEAST(major, minor) \
+    (__GNUC__ > (major) || (__GNUC__ == (major) && __GNUC_MINOR__ >= (minor)))
+#else 
+# define GCC_AT_LEAST(major, minor) 0
 #endif
 
-#ifdef __SSSE3__
-#include "tmmintrin.h"
+#ifdef __clang__
+#ifdef __has_attribute
```
marcelm, line 10: Since the above macro uses `#if defined` instead of `#ifdef`, can you do that here as well for consistency? And the two `#ifdef`s can then also combined into `if defined ... && defined`.

src/dnaio/bam.h
```diff
@@ -1,21 +1,26 @@
...
+#define CLANG_COMPILER_HAS(attribute) __has_attribute(attribute)
+#endif 
 #endif
+#ifndef CLANG_COMPILER_HAS 
+#define CLANG_COMPILER_HAS(attribute) 0
```
marcelm, line 15: This can then be in an `#else` branch

rhpvorderman:
Thanks for the review. It is done. Also I found some issues when compiling on ARM64 when this code was incorporated into sequali. I have updated the macros such that they will work now on ARM64 as well. 

I am not very fond of all this compiler checking macors, but I see no better way alas.

PR #133, opened 2024-04-08
Rewrite ascii check to allow compiler auto vectorization

rhpvorderman:
Rewriting it to use `size_t` will create the most optimal code for all platofrms. Also I did a small optimization that works well for small strings. 

Say if the string is length 13. Previously the function would check 8 bytes at once, then check 5 bytes individually. What I do now is simply substract 8 from the length and check an 8 byte vector at the end. This is much more optimal for smaller strings. 

Not that we need that, as we check 128kb chunks. But it eliminates a loop from the function, so might as well leave it in. 

The 8byte loop is unrolled to do 4 OR operations simultaneously. This is automatically compiled to do two SSE2 OR operations with `-O2` on GCC. As a result this code is not slower than the code that it replaces with had vectors inserted manually. 

The resulting code is much more portable as a result, and should also compile to use vectors on ARM platforms.

EDIT: See also [godbolt compiler explorer](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJmkOAMngMmABy7gBGmMQgkgCspAAOqAqEtgzObh7e0onJNgIBQaEsEVGxFphWeQxCBEzEBOnunj7llak1dQQFIeGR0XEKtfWNmS2Dnd1FJf0AlBaorsTI7BwApF4AzIHIblgA1KsbjoPoWFQAdAiH2KsaAIK3D5tngZh7d0KOAJJfxgCyHwA0sYABwAIQAmgAVbB7DSqEEaRHIpGolFyPx%2BR7rDYvILvT4/f5A4xcSEwuEIjTY%2B7jGzIPaBWivRmCPaDYiBYDGPAKYxMBTIPB4YxUMS0MJMZAAawgaAYgz2yAQdT2ACoOVzSOy8AAvTDGAh7ehGAgIGbYgDsYMeeztOv1hr24uMyrqCgOGwAInDDjb7va9vxiHsIMlHUa8IcvdSNmDGYdHCbgGa/QmvDaMzMDtbbYH7S63cQPatLY5o%2ByCJyjKsYmCozEvX683bS02A/biJgCIsGHswJBCyriwcvAA2AnfX4AoTAsnQ7AWuNW9sPe6BI2aow8vkCoV4OUCRVF9Vb4Da8MGo3Js3Zq3%2Bu6BvBUUM3hCexwOzCoKhhvVXmY71zDt8z2Lse2IPszx3flBWFUVxUlGUwyrLVjUYFNzWbECc1XQNLydBhemIYwf1dBBXAYaUS29dDTXfAB6L8fz/CMlwffD/ydNVlUo6jPR9Vir3VGYzzTFsv0I4jSKoYxKOIFxaBon0iOKSIZPIviPSYyRsMfe0CKNLsWCYQIuU0qjlL2VSSg03jLL2ABaUMbPUsj5MUj01T2SR2Ikwz1Xs6VjHiKsBKVCjLL0ziI2dWhaHI90NHC2MOIMrijSHd0uBS6L0tirLiy8XLlxwgLCoUDYSrSu1g1DAKo29VLGQ/azpPchgFPihQ0yjDN1jBICavzCrktLctaKCkKq1rMFqUbPLQIqnLxorKbQuIWauFrJtSv0pb4sSoqcwmn11pmusvB2xaRsOotKpOtbIuCjbZo2a69tAu1zpDAaK10z77TbfyMrihL7vC0aTrBo6FBWssYfu4rxsR4dKpuuqhKdRqY16xNjNMhhzKCnq43TTNBpzYbAwqx7Jue6bNrrBtduG4GcIY7y/FQJh0D2M03gMRUqEwAB3PYwgATwITAPXoKhIz7JgdSMehWRl4BIiDEggwFI1UHiSImCqBRzgOHD7ShBBIjeXl2VQNg9lQAA3SIDHiPZra7PZRcId8BZ9khpT2dABDeCJg0wbUwlcI06uV5IWHiNWJILOC8AizAZX5hA7dD2XrNQI0TIIGXiDNtUGJB2KhYICzg4rNUsaNNUZhQ6tgFHeM3yc5jf0MwCbtp1baNr%2BubvA3t%2B0HO60dHCcPinYlZ1BclFz09m7g4OZaE4GJeE8DgtFIVBOE/BQFiWN4cR4UgCE0be5mlEAYg0fROEkA%2BH5PzheAUEA373yPtvUgcBYBIDQEnOgkRyCUEgfEaBURkDAC4BsLwfA6Bl3/hAMI38wiBDqJLTgt98HMGIJLAA8mEbQWcgG30gWwQQFCGC0CIcA0gWAY7AEcOKf%2B3BeBYBMqrFYx98BdmsHgV2fDj6YFUFnWOKxb4bgqN/ZkYRiCEOcFgb%2BqEWDEN4K7YgYQkiYC9JgIRwBmRGAfnMKgBhgAKAAGp4DFhQw2h9b78EECIMQ7ApAyEEIoFQ6h2G6C4PoQwJgzD6DwGEf%2BkA5gGxNpwRyFCvC8BdpETkWB4kQDmJYWhqR7AMCcC4Joeh/CvCmH0cJOQUgCBGJ4WpSR6kMEmMRPQBSJECA6MMMpmRwldKqL0roVSOmDKGA0fpTSLCTPaWpKIXB8mX2WBIHee8v7sNPhwPYCIxyOTHJIPYwBkAMlQecYqEBcCEB1jfGYvAgFaEAqQZ%2Br934cE/qQQ%2Bx9tl/wAXfGx6yODpK%2Bd/X5ALgHPMMckOwkggA%3D%3D%3D)

### DIFF
--- a/src/dnaio/ascii_check.h
+++ b/src/dnaio/ascii_check.h
@@ -1,11 +1,17 @@
 #include <stddef.h>
-#include <stdint.h>
-#ifdef __SSE2__
-#include "emmintrin.h"
-#endif
 
 #define ASCII_MASK_8BYTE 0x8080808080808080ULL
 #define ASCII_MASK_1BYTE 0x80
+
+static inline int string_is_ascii_fallback(const char *string, size_t length)
+{
+    /* Combining all characters with OR allows for only one bit check at the end */
+    size_t all_chars = 0;
+    for (size_t i=0; i<length; i++) {
+        all_chars |= string[i];
+    }
+    return !(all_chars & ASCII_MASK_1BYTE);
+}
 
 /**
  * @brief Check if a string of given length only contains ASCII characters.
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Damn! Clang does not recognize the optimization opportunity. But it does so for the last loop (that is not unrolled). However if I write it like that, GCC does not recognize the optimization opportunity. In any case, this code + GCC optimization results in the most optimal assembly. But it will be not usable for clang users.

So, best to not merge this, I guess?

rhpvorderman:
@marcelm, friendly reminder ping. I decided to merge this code into sequali because GCC is used on linux and Linux is the only production platform for bioinformatics. (I am not aware of any MacOS and Windows compute clusters out there). Furthermore it should be a lot faster on platforms that do not benefit from auto vectorization due to the loop unrolling. In addition to that, I think the code is nicer when it does not make have use of `if __SSE2__` and other compile guards. But it is your call in this case.

PR #135, opened 2024-06-30
Decode BAM sequence with less code. Dispatch function at dynamic link time rather than at run time. 

rhpvorderman:
Two improvements coming from htslib. Credit for simplifying the bam decode goes to @jmarshall. 

See: https://github.com/samtools/htslib/pull/1795

For the use of constructor:
https://github.com/samtools/htslib/pull/1786

Using the constructor attribute makes the code simpler. Another change is that the fallback code now also always use a function pointer and cannot inline. This also makes the code simpler.

### DIFF
--- a/src/dnaio/bam.h
+++ b/src/dnaio/bam.h
@@ -72,49 +72,25 @@ decode_bam_sequence_ssse3(uint8_t *dest, const uint8_t *encoded_sequence, size_t
     const uint8_t *dest_end_ptr = dest + length;
     uint8_t *dest_cursor = dest;
     const uint8_t *encoded_cursor = encoded_sequence;
-    const uint8_t *dest_vec_end_ptr = dest_end_ptr - (2 * sizeof(__m128i));
-    __m128i first_upper_shuffle = _mm_setr_epi8(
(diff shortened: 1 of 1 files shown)


rhpvorderman:
@marcelm, friendly reminder ping. This exact code is already merged in htslib. I hope you had a nice holiday!

marcelm:
Still not really working fully, but if it’s identical to what is in htslib, I’ll just merge it.

rhpvorderman:
> Still not really working fully

I don't understand, can you elaborate?

marcelm:
I am still on vacation and don’t systematically do work-related things that are on my to-do list, only some things that I feel like doing. I’ll be back at work middle of August.

rhpvorderman:
Oh sorry. I saw a github reply somewhere and assumed you were back. I won't ping you anymore before the middle of August.

marcelm:
Yes, I thought so (that you saw something from me), which is why I shouldn’t have done anything at all ;-).

rhpvorderman:
Well now that I am in the know I won't do it anymore ;-). Fun fact: I wrote this code during my holiday. Writing assembly in the form of intrinsics is simply too much fun.

PR #141, opened 2024-10-08
Allow chunking uBAM

rhpvorderman:
This allows reading raw BAM records with no header present. This format cannot be detected so must be provided beforehand. "bam_no_header" was chosen as the format name. 

A chunker is also provided that can provide chunks of BAM records. These can be read using the "bam_no_header" format in dnaio.open. 

With these additions cutadapt should be able to set up multithreaded reading of BAM records.


rhpvorderman:
@marcelm  I did a prove of concept here: https://github.com/marcelm/cutadapt/pull/812/files 

It is possible to get multithreaded BAM reading going with just the changes in this PR.

In the future we need to think how to handle the header. Ideally the header is copied from the input tag with an additional PG line added from cutadapt.

src/dnaio/_core.pyx
```diff
@@ -802,6 +818,7 @@ cdef class BamIter:
                 self._read_into_buffer()
                 continue
             record_size = (<uint32_t *>record_start)[0]
+            print(record_size)
```
marcelm, line 821: Leftover from debugging?

marcelm:
Awesome, thanks! Will hopefully be able to review this towards the end of the week, no time today.

src/dnaio/_core.pyx
rhpvorderman: Yes, exactly. Thanks for spotting.

rhpvorderman:
@marcelm . Friendly reminder ping. I am currently watching cutadapt blazing through some nanopore files at only one core, which reminded me of this PR. Please do not take this as pressure. I am happy to wait for a few weeks more.

src/dnaio/_bam.py
```diff
@@ -0,0 +1,43 @@
...
+    magic = fileobj.read(4)
+    if not isinstance(magic, bytes):
+        raise TypeError(
+            f"fileobj {fileobj} is not a binary IO type, " f"got {type(fileobj)}"
+        )
+    if len(magic) < 4:
+        raise EOFError("Truncated BAM file")
+    if magic[:4] != b"BAM\1":
+        raise ValueError(
+            f"fileobj: {fileobj}, is not a BAM file. No BAM magic, instead "
```
marcelm, line 15: In the user-visible messages, I’d call it "file signature" instead of magic.

marcelm: Ah ok, I saw this is just moved code. Well, feel free to change it or not.

marcelm (approved):
Awesome, this looks good! And sorry for the delay. It did not actually take that long to review, should have done that earlier.

So one thought I had is that this way of chunking only works for single-end reads. So that is one more issue that would need to be solved in order to enable BAM paired-end input.

With Python’s support for free threading coming, I’d actually hope that we could get rid of all this chunking business in the future. Splitting the input up into chunks of bytes is only necessary because sending already parsed records from one process to another using multiprocessing is expensive. With threads, we could have one thread that is dedicated to parsing the input and supplying worker threads with lists of records. The parser thread can then make decisions based on the content of the records, that is, it can ensure that a worker always gets both paired-end reads.

marcelm:
Feel free to merge when ready.

rhpvorderman:
Done. Thanks for the review!

PR #145, opened 2024-10-15
Ignore complex mypy issue

rhpvorderman:
The original error:

```
src/dnaio/multipleend.py:205: error: Argument 1 to "enter_context" of "ExitStack" has incompatible type "Any | str | PathLike[Any] | BinaryIO"; expected "AbstractContextManager[IO[Any], bool | None]"  [arg-type]
Found 1 error in 1 file (checked 12 source files
```


rhpvorderman:
Thanks! I got a lot of "build failed" spam from github because of this while I am working on the tag support.

PR #148, opened 2024-11-11
Fix failing test on Windows with Python 3.11.9+

marcelm:
Python 3.10 is fine, but 3.11 breaks the tests. This made creating the wheels for v1.2.3 fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,16 +49,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         include:
         - os: macos-13
           python-version: "3.10"
         - os: macos-14
-          python-version: "3.10"
-        - os: windows-latest
-          python-version: "3.10"
-        - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


rhpvorderman:
1.2.2 Was released but a month ago and the wheel building presumably went fine (I saw no error messages in the e-mail). I wonder what could cause this.

rhpvorderman:
Could it be this change? https://github.com/python/cpython/pull/116989/files

It is the only thing I find in the changelog (https://docs.python.org/release/3.11.9/whatsnew/changelog.html#python-3-11-9) that seems related.

rhpvorderman:
Alternatively, I wonder if replacing `cat` with `python -c 'import sys; file = sys.argv[1]; f=open(file, "rt"); text = f.read(); f.close(); sys.stdout.write(text)` would solve the issue. That should work with all operating systems equally.

marcelm:
Or maybe this one? https://github.com/python/cpython/issues/95782

In any case, the problem is in `_detect_format_from_content`, where `seekable()` is true for stdin on Windows but doesn’t actually work. Not sure what the actual problem is, but preferring `peek()` makes the issue go away.

rhpvorderman:
Good find! These windows-specific issues are always very tedious to solve. I am glad you find the solution relatively quickly.

marcelm:
Version 1.2.3 is now released.

PR #150, opened 2025-04-07
Use an SPDX license expression

marcelm:
This avoids a setuptools warning.

See PEP 639


rhpvorderman:
Readthedocs fails but it is unrelated.

